PR #7, opened 2020-09-16
device refactor

ajbalogh:
Refactor of #/components/schemas/Device... schema objects.
Removed ixia-isms and scaling features.


SubhajitPalKeysight (commented):
Hi Andy ,  looked into the new model and we also need to deliberate  how this model fits into ixnetwork as well :

> This model may have a scaling issue –

-Think about the script logic we need to write/call api’s N number of times to create devices/stack/protocols – script complexity will increase.
-End point mapping in traffic source and destination will be more complex.
-IxNetwork concrete implementation on per devices  will not scale ( in ngpf )  as configuration will end up creation of many devices and layered stack.

> Complex scenario - like EVPN ( DG -> NG -> DG/Loopback/BGP -> VRF/Mac/)  we need unconnected interface , there should be some container top of port
> Number of Devices removed : Needed to map Number of IP/stacks ( emulation of multiple IP's ) 
> flow/endpoint.yaml : Meshing is needed
>Traffic end point – device is missing 


Thanks
subhajit

device/device.yaml
```diff
@@ -8,31 +8,22 @@ components:
...
-        devices_per_port: 
+        port_name: 
           description: >-
-            The number of emulated devices that will be created on each port.
-          type: integer
-          default: 1
-        ethernets:
+            The name of a port that will host the emulated device stack.
+          type: string
+          x-constraint: '#/components/schemas/Port.Port.properties.name'
```
winstonliu1111, line 22: you mentioned Port.Port -> Port, does that apply to this x-constraint?

device/device.yaml
```diff
@@ -8,31 +8,22 @@ components:
...
-            $ref: '#/components/schemas/Device.Device'        
-        networks:
-          description: >-
-            An unordered list of 0..n emulated device containers.
-          type: array
-          items:
-            $ref: './network.yaml#/components/schemas/Device.Network'        
+            The type of top level emulated protocol device. 
+          type: string
+          enum: [ethernet]
```
winstonliu1111, line 27: it's more than ethernet for the enum?  or are those basic/initial set of choices coming in another PR?

winstonliu1111 (approved):

layer1/layer1.yaml
```diff
@@ -65,9 +67,21 @@ components:
...
+      description: >-
+        Container for virtual ethernet settings
+      type: object
+      properties:
+        speed:
+          description: >-
+            This is the speed that will be set on the virtual ethernet interface.
+          type: string
+          enum: [one_hundred_mbps, one_gbps, ten_gbps, twenty_five_gpbs, fity_gbps, one_hundred_gpbs]
+          default: one_gbps
```
ankur-sheth, line 80: We might want the default to be ten_gbps.  @ashutshkumr, thoughts?

device/device.yaml
```diff
@@ -6,33 +6,25 @@ info:
...
-            $ref: '#/components/schemas/Device.Device'        
-        networks:
-          description: >-
-            An unordered list of 0..n emulated device containers.
-          type: array
-          items:
-            $ref: './network.yaml#/components/schemas/Device.Network'        
+            The type of emulated protocol device. 
+          type: string
+          enum: [ethernet, vlan, ipv4, bgpv4]
```
ankur-sheth, line 20: Is this missing ipv6?

device/ipv6.yaml
```diff
@@ -21,4 +21,6 @@ components:
           $ref: './patterns.yaml#/components/schemas/Device.Pattern'
         manual_gateway_mac:
```
ankur-sheth, line 22: I thought that you had removed this?

ankur-sheth (approved):
Added a few minor comments in various places.

ajbalogh:
I want to point out to everyone based on conversations with Matt/Winston/Ankur that I changed the `Endpoint` naming convention to just SrcDst for now.  

That means you can assign source/destinations using either a PortSrcDst container or a DeviceSrcDst container to a flow.
Please check the descriptions on those schema objects for the differences between them. 

If there are any better suggestions for those names please feel free to suggest!

This sample illustrates how to use them using the new device models.

```python
tx = Port(name='tx')
tx.devices.append(Device(name='txd', choice='ethernet'))
rx = Port(name='rx')
rx.devices.append(Device(name='rxd', choice='ethernet'))

flow = Flow(name='test', src_dst=SrcDst(PortSrcDst(tx_port_name=tx.name, rx_port_names=[rx.name])))
flow = Flow(name='test', src_dst=SrcDst(DeviceSrcDst(tx_device_names=[tx.devices[0].name], rx_device_names=[rx.devices[0].name])))

```

device/device.yaml
```diff
@@ -6,33 +6,25 @@ info:
...
-          type: array
-          items:
-            $ref: './network.yaml#/components/schemas/Device.Network'        
+            The type of emulated protocol device. 
+          type: string
+          enum: [ethernet, vlan, ipv4, ipv6, bgpv4]
+        ethernet:
+          $ref: './ethernet.yaml#/components/schemas/Device.Ethernet'
+        vlan:
+          $ref: './ethernet.yaml#/components/schemas/Device.Ethernet'
```
winstonliu1111, line 24: Device.Vlan?  same comments for the ipv4/ipv6/bgpv4

result/capture.yaml
```diff
@@ -0,0 +1,42 @@
+
```
winstonliu1111, line 1: what is the expectation of capture result? what is the format of the returned packets?  i wonder if it is sufficient to get back a link to pcap, so that the client can use other tools to dissect/interpret as they wish?

api/api.yaml
```diff
@@ -119,7 +139,31 @@ paths:
...
+        - Results
+      operationId: get_capture_results
+      requestBody:
+        description: >-
+          Capture results request to the traffic generator.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../result/capture.yaml#/components/schemas/Result.CaptureRequest'
```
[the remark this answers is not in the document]
ajbalogh: Result requests are currently separated from generic results.  I could rename these files but they are just temporary as they get bundled into a single openapi.yaml file.

result/capture.yaml
ajbalogh: That's a good point - but should it be a two step process with specific transport implications (ie http link) or just return the pcap as application/octet-stream (binary). Also the start/stop is on multiple ports, the get results is on multiple ports - so the structure being returned can be array of port name and pcap binary (array of objects allows for model growth with no bw compat issues)

ajbalogh:
Updated capture API to follow how a user would actually capture. Typically a user starts a capture tool on an interface with/without filters. After a period of time the stop it and get the results. The capture API's are changed to follow that model, /control/capture starts a capture on a specific port and /results/capture stops the capture and returns the captured file as binary so that it can be saved to a file.

layer1/layer1.yaml
ashutshkumr: Either one of `one_gbps` and `ten_gbps` is fine IMO.

ashutshkumr:
> I want to point out to everyone based on conversations with Matt/Winston/Ankur that I changed the `Endpoint` naming convention to just SrcDst for now.
> 
> That means you can assign source/destinations using either a PortSrcDst container or a DeviceSrcDst container to a flow.
> Please check the descriptions on those schema objects for the differences between them.
> 
> If there are any better suggestions for those names please feel free to suggest!
> 
> This sample illustrates how to use them using the new device models.
> 
> ```python
> tx = Port(name='tx')
> tx.devices.append(Device(name='txd', choice='ethernet'))
> rx = Port(name='rx')
> rx.devices.append(Device(name='rxd', choice='ethernet'))
> 
> flow = Flow(name='test', src_dst=SrcDst(PortSrcDst(tx_port_name=tx.name, rx_port_names=[rx.name])))
> flow = Flow(name='test', src_dst=SrcDst(DeviceSrcDst(tx_device_names=[tx.devices[0].name], rx_device_names=[rx.devices[0].name])))
> ```

@ajbalogh @ankur-sheth @winstonliu1111 @SubhajitPalKeysight @SuvenduMozumdar 

I may be stepping into something very subjective; I like the idea of `src` / `dst` as attribute name (for flow origin / sink) but I think we'll be more consistent overall if we just accept `name`s as values. 

```python
f1 = Flow(name='f1', src='p1', dst='p2')
f2 = Flow(name='f2', src='e1', dst='e2')
# or to be more descriptive
f1 = Flow(name='f1', src_name='p1', dst_name='p2')
```

Once we do that, we can extend or evolve over time what can be or cannot be a flow `src` or `dst`.
i.e. As of now `src` or `dst` can be:
- Test port (Ixia ports or OS/DPDK interfaces)
- Group of test ports (e.g. LAG, multiple destinations, etc.)
- Emulated devices pertaining to any supported encapsulation (e.g. 1 emulated device x 1 test port)
- Group of emulated devices distributed across group of test ports (e.g. M emulated devices x N test port)

Now, coming back to models that describe `src` and `dst`, I think the terms `Endpoint` or `Node` offer better understanding of underlying entity over `SrcDst`. 

Let's say we pick one anyway, e.g. `Node`, there are two approaches to defining them:
- Different models for different kinds, e.g. `PortNode`, `DeviceNode`, etc.
- Same abstract model called `Node` which is composed of `PortNode`, `DeviceNode`, etc.

There's also a realization that `Emulated Devices` are eventually composed of test ports + device-specific configuration; or more generally, a `Node` can be composed of other `Node`s + specific configuration. i.e. Test ports and emulated devices could possibly be a part of same model hierarchy !

For first approach, current spec shall suffice. But the benefit of using second approach is that:
- All the src/dst `name`s would belong to same model's namespace (I'll expand on this further if required).
- We can leverage inherent relationship between various kinds of `Node`s to build more complex test topology

Unrefined example:
```python
# single test ports (empty base_nodes by default)
_ = Node(name='p1', choice='port', port=PortNode(...))
_ = Node(name='p2', choice='port', port=PortNode(...))
# group of test ports
_ = Node(name='p3', choice='port_group', base_nodes=['p1', 'p2'])
# emulated ether device (scaling is part of ethernet node config)
_ = Node(name='e1', choice='ethernet', ethernet=EthernetNode(...), base_nodes=['p1'])
# emulated ipv4 device built on top of emulated ether device (scaling is part of ipv4 node config)
_ = Node(name='e2', choice='ipv4', ipv4=IPv4Node(...), base_nodes=['e1'])

# stateless flow
_ = Flow(name='f1', src='p1', dst='p2')
# stateless flow with one source and multiple destinations
_ = Flow(name='f2', src='p1', dst='p3')
# ipv4 flow (where MAC / IP info is learned)
_ = Flow(name='f3', src='e2', dst='e3')
```

P.S. `Endpoint` or `Node` don't mean the same thing, so feel free to `sed "s/Node/Endpoint/g"` in example above.

result/capture.yaml
```diff
@@ -9,34 +9,35 @@ components:
...
     Result.CaptureRequest:
       description: >-
-        The capture result request to the traffic generator 
+        The capture result request to the traffic generator.
+        Stops the port capture.
       type: object
+      required: [port_name]
       properties:
-        port_names:
+        port_name:
```
winstonliu1111, line 17: is there a reason now this only stops a single port instead of list of ports?  same question as above for Control.PortCapture..

result/capture.yaml
```diff
@@ -9,34 +9,35 @@ components:
...
+      properties:
+        port_name:
+          description: >-
+            The name of the port the capture was taken on.
+          type: string
+        format:
+          description: >-
+            The requested format of the capture file.
+          type: string
+        capture:
```
winstonliu1111, line 37: @ankur-sheth Do we serialize this ourselves?  or is there a way to provide a link to the capture file that the user can retrieve separately?  I guess we should probably brainstorm a bit on what is the ux here?  problem with providing a link is that we don't know how long we need to keep it on the server side.

result/capture.yaml
ajbalogh: I envision taking a capture as two steps - much like you do when you use tcpdump, start capturing (provide all details such as filters etc), then stop capturing (return the capture).  
No saved files, no special file locations, no special filenames (special characters, linux/windows), no download links, download format (multipart, octect-stream) etc.

winstonliu1111: i guess the trouble i have reconciling is what is the expected output to the user if say, they started a capture at time x for port abc, and say issue a results for port abc's capture at time x+ 1 hour later.  I'm just wondering as a user, what I'd expect to be in that response for the results....just seems we can always run into some kind of resource limits on the TG side of things pending how much is captured.

ajbalogh: Resource limits would be implementation specific. Since the proposed schema of capture start is an object it can be extended with additional properties as needed (can add a capture limit, stop after x seconds etc) rather than trying to figure out all of those details now.  The current modeled API is start capture (on x number of ports with optional filters). Get a port's capture when the need arises. Implementations can cap however they want and expose that in the /results/capabilities API.

The sample illustrates starting a capture on a port, getting the capture and writing the capture to a file.  This leaves the file name and the file location up to the user. 
```
api.set_port_capture({'port_names': ['p1'], 'filters': null, 'format': 'pcap'})
capture = api.get_capture_results({'port_name': 'p1'})
with open('p1_%s.cap' % time.time(), 'wb') as fid:
    fid.write(capture)
```

winstonliu1111: i guess what i meant is there will be implementation specific cap and the question is how would a user know whether in the example code above whether the received data spans the first 5 minutes or the last 5 minutes of the capture, assuming time between start capture and results is say 1 hour, and the target only has a 5 minute worth of buffer (assuming some rate)?  is there a way to maybe signal some metadata back assuming we hit some kind of cap in the returned result?  Sorry, i'm asking all these w/o a real use case in mind....maybe overthinking it.

ankur-sheth: Agree.  I have some of these same questions in mind and it is not clear to me which way things should go.  My initial reaction was that I think of stopping and getting the capture as two distinct steps and I didn't like the fact that start is an action under 'control'  but stop is an action under 'results'.  When I think about it further, it isn't that obvious to me.   And I still have the same questions about limits and stuff how things would behave with those constraints.  Need some solid use-cases in mind.

ajbalogh: i thought pcap has timestamps?

Time stamp usage considerations
Now, there are several important aspects that every network forensic specialist needs to be aware of. Firstly, timestamps are derived from the clock on the machine performing the packet capture. Therefore it’s important to check the accuracy and record this machine’s system time before starting the packet capture.

Secondly, time stamps are saved to the .pcap file in GMT/UTC format, which can easily be verified by examining the value of “thiszone” in the global header. For display purposes, packet capture software such as Wireshark automatically converts these timestamps into the local time of the machine used to examine the .pcap files. However UTC times can also be displayed in Wireshark by adjusting its “Column Preferences”.

ajbalogh: I'm trying to avoid having information in the response that is duplicated - such as timestamps which is already included in the pcap.
Limits (and any other request information) can be added to /control/capture request since its an object - i just don't want to add it right now since there is no specific use case for it.

winstonliu1111: I'd assume the RDMA use case right now is setting up traffic in a way where it is probably a fixed amount that one would know it will never exceed any practical resource limit on the device performing the capture?  So their use case should be okay.  I guess my question is do we want to get the implementation to return a signal to the end user indicating whether the capture may be truncated or not (due to resource limit, whether a capacity limit in the capture buffer, or it has wrapped a circular buffer), just a simple true/false..  This is so that if it comes back as "true", the user at least have a definitive hint on whether he needs to put in better capture filters, or set up the traffic in a way where he won't exhaust the resource limit on capture.

layer1/layer1.yaml
```diff
@@ -65,9 +67,21 @@ components:
...
+    Layer1.EthernetVirtual:
+      description: >-
+        Container for virtual ethernet settings
+      type: object
+      properties:
+        speed:
+          description: >-
+            This is the speed that will be set on the virtual ethernet interface.
+          type: string
+          enum: [one_hundred_mbps, one_gbps, ten_gbps, twenty_five_gpbs, fity_gbps, one_hundred_gpbs]
```
winstonliu1111, line 79: can we do 100mbps, 1gbps, 10gbps, 25gbps, 50gbps, 100gbps (btw, just noticed typo on fity_gbps and one_hundred_gpbs)?

result/capture.yaml
ajbalogh: The OpenAPI standard allows for streaming and non streaming responses.
A non-streaming response will be available for those implementations that can only return a capture file when requested.
A streaming response will be available for those implementations that do not want to incur storage overhead and put the onus on the user to interpret/store/stop the stream of capture results.

layer1/layer1.yaml
ajbalogh: We should switch all enums that have numbers to that model then.

ajbalogh: and make it part of the modeling guide

ashutshkumr: Please note, enum names starting with numbers are not very portable across languages (i mean sure this would pass code generation, but it may transform the names to something unwanted).

SubhajitPalKeysight (changes requested):
Bidirectional and fully mesh across source and destination is also needed. Many sonic test cases has that. Model should expose such knobs so that user can build up script relatively easily.

layer1/layer1.yaml
ankur-sheth: Do you have examples?  Which languages?  If it is breaking in some language(s) that we don't care about then, we could ignore it.

winstonliu1111 (approved):

result/capture.yaml
winstonliu1111: just occurred to me, for streaming response, how does the user stop the capture?

layer1/layer1.yaml
ashutshkumr: Summary: bad for *.proto.

Works fine with Python (since enum in python is a string which is validated on client side).
And Go too for similar reason (except enum validation doesn't work so they're reconsidering other approach: https://github.com/go-swagger/go-swagger/issues/861)

On protobuf, this is the generated code (note leading underscore). When code is generated from this, the field names will be further transformed into something without leading underscore (https://github.com/golang/protobuf/issues/201).
```protobuf

  // This is the speed that will be used if auto_negotiate is false.
  enum speed {
    _1000_MBPS = 0;
    _100_FD_MBPS = 1;
    _100_HD_MBPS = 2;
    _10_FD_MBPS = 3;
    _10_HD_MBPS = 4;
  }
```

ajbalogh: This should be reverted to the original.

winstonliu1111 (approved):

winstonliu1111 (approved):

ankur-sheth (approved):

device/device.yaml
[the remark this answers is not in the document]
ajbalogh: At this time I am holding off on adding bidirectional and any kind of meshing as this can be accomplished by adding more flows. If this proves to be a problem an separate issue can be opened.

PR #10, opened 2020-09-24
async response proposal

ajbalogh:
This is a standard async response proposal that servers can implement.


winstonliu1111 (approved):
LGTM.  are we adding the redirect status in this PR as well?

ajbalogh:
The status codes in the request's response are independent of the configuration state.  
They exist to give the user an indication of the state of their request.
For example the server can be applying a configuration but a subsequent transmit request should fail with a 409 error indicating the server is busy with a previous request. 

To reiterate everything in the API/schemas is optional - an implementation can choose what to support, so in the case of POST /config response codes athena can support 204, 400, 409 and indicate in /results/capabilities that 202, 300 is not supported (ie no progress polling, use /results to determine when to issue the next request).

winstonliu1111 (commented):
@ankur-sheth @ashutshkumr we should probably sync up with andy, after few days of thinking about it, i feel the current proposal on async is too tied to http as a transport, so we're sort of not agnostic to transport any more.

winstonliu1111 (changes requested):
we should probably review the transport dependency again.

ajbalogh:
By modeling with openapi it implies using rest(http) but it shouldn't constrain an implementation - ie athena is implementing the model using go on the server and python on the client.

The use of error codes is used to differentiate between responses to a request but it isn't the only thing  that differentiates between the responses.

For example - set_config returns two successful responses - in python I would call set_config and receive a success response object (modeled as a 204 no content response) or an async response object (modeled as a 202 content response).
It also returns an error response object (400) or a conflict response object (409 - ie you tried to set config while an existing set_config is being done).

The code generator if its any good should generate a docstring for the set_config method (the information should come from the response schema objects) or if its not then the implementation will have a samples library or their own documentation.

```
def set_config():
    """The set config operation

    Returns:
    - obj(SuccessNoContent)
    - obj(SuccessAccepted): poll using progress_url, cancel using cancellation_url
    Raises:
    - obj(BadRequestError)
    - obj(RequestConflictError)
    """"
```

ajbalogh:
I still haven't finished the /results/state responses - those will come later as required.
I will be posting sample code later today.

ajbalogh:
sample code using the proposed model

```

# configuration definition
config = [
    ports = [
        {
            'name': 'p1'
        }
    ]
    flows = [
        {
            'name': 'f1',
        }
    ]
]

# set the traffic generator configuration state
config_state = {
    'choice': 'config_state',
    'config_state': {
        'state': 'update',
        'config': config
    }
}
# any 409, 590 responses will be raised as exceptions
api.set_state(config_state)

# set the traffic generator flow state
flow_transmit_state = {
    'choice': 'flow_transmit_state',
    'flow_transmit_state': {
        'state': 'start',
        'flow_names': [config['flows'][0]['name']]
    }
}
# any 409, 590 responses will be raised as exceptions
api.set_state(flow_transmit_state)
```

api/api.yaml
```diff
@@ -16,130 +16,45 @@ servers:
...
-        Use the /results/capture API to stop a port capture and get the captured
-        results.
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '../control/control.yaml#/components/schemas/Control.PortCapture'
-      responses:
+          $ref: '../result/request.yaml#/components/responses/Pending'
```
winstonliu1111, line 33: are we removing 202 for now based on this morning's discussion to keep it synchronous?

api/api.yaml
```diff
@@ -16,130 +16,45 @@ servers:
...
       responses:
-        '204':
-          description: >-
-            No content
+        '200':
+          description: List of states
+          content:
+            application/json:
+              schema:
+                type: object
```
winstonliu1111, line 51: free form objects?

ajbalogh:
@winstonliu1111 @ashutshkumr @ankur-sheth  just waiting for your reviews before I merge.

winstonliu1111 (approved):

control/control.yaml
```diff
@@ -8,15 +8,52 @@ info:
...
+          type: string
+          enum: [config_state, port_link_state, port_capture_state, flow_transmit_state]
+        config_state:
+          $ref: '#/components/schemas/Control.Config'
+        port_link_state:
+          $ref: '#/components/schemas/Control.PortLink'
+        port_capture_state:
+          $ref: '#/components/schemas/Control.PortCapture'
+        flow_transmit_state:
+          $ref: '#/components/schemas/Control.FlowTransmit'
```
ashutshkumr, line 27: Can we rename,
- flow_transmit_state -> flow_state
- ControlFlowTransmit -> FlowState

Reasons:
- `Result` refers to it as FlowState
- Technically, the overall effect is not limited to transmit state

result/request.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+info:
+  title: Request response models
+  description: >-
+    All standard request response schemas
+  version: ^0.0.0
+
+components:
+  responses:
+    SuccessNoContent:
```
ashutshkumr, line 11: We did have lengthy discussions after which we came up with sync approach (which can be later extended to async approach with little modification).

But, it's still not clear why we stuck to 204 instead of 200 ? One very important scenario (among numerous others) where returning response body makes sense is when we need to return warning. e.g.
- User configures a flow with Ether/VLAN/IP/TCP and gave a frame size 64B (because that's a valid scenario)
- But internally, we might need to add instrumentation, which will `require adjusting the size to >64B`
- In response, user doesn't care about the internals, but `must be` informed about the adjustments in form of a warning (otherwise they'll just be confused looking at the increased frame size in results)

ashutshkumr (changes requested):
Apart from aforementioned modifications, we also need to add descriptions for `results/*` endpoints.

ashutshkumr (approved):
LGTM, but following minor items still need your attention.

- Regarding https://github.com/open-traffic-generator/models/pull/10#discussion_r499157746, I meant that success response should contain `warnings` (different from `errors`).
- Regarding https://github.com/open-traffic-generator/models/pull/10#discussion_r499155056, I was suggesting renaming `FlowTransmitState` to `FlowState` (details in the comment)

PR #12, opened 2020-09-30
Proposal for alternate asynchronous control implementation (actions and status)

ashutshkumr:
What's new ?
- Rephrased purpose of POST /config: It will validate the Config and store it (and not actually apply it).
- Added GET /config: It will return Config per last successful POST.
- Added POST /control/action: For invoking a list of actions (mix of granular and composite actions)
- Added GET /control/status: For checking statuses of previously invoked actions

Motivation ?
- We should be agnostic of underlying transport and be more model driven
- Why can't we represent statuses of all possible controls just as a stat ? e.g. when traffic run is in progress, users know they need check port or flow stats to see the progress. Likewise, when control operation (or action) is in progress, users should check control stats (or status) to see the progress.

Example:
```python
from athena import *


config_api = ConfigApi()
control_api = ControlApi()
cfg = Config(name='test', ports=[Port(name='p1')], flows=[Flow(name='f1')])

# push configuration
errs = config_api.set_config(cfg)
# check for validation errors
assert len(errs.errors) == 0

# start traffic run
errs = control_api.set_control_action(ControlAction(actions=['start']))
# check for validation errors
assert len(errs.errors) == 0

# wait for traffic run to be started
while True:
    status = control_api.get_control_status()
    if status.start == 'done':
        break
    if status.start == 'failed':
        print("Traffic run failed: %s" % status.errors[-1].message)
        break
```

What's pending ?
- Currently, to implement tri-state (none, failed, done), each ControlStatus attribute has its own ENUM definition. I'd like to define ENUM separately and reuse it.
- Any suggestions for naming convention to be used for granular control actions ?
- We might want to incorporate some delay between each consecutive actions. How to deal with that ?

### DIFF
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -24,7 +24,7 @@ paths:
         - Config
       operationId: set_config
       description: >-
-        Sets the configuration on the traffic generator.
+        Creates new or updates existing traffic generator configuration.
       requestBody:
         required: true
         content:
@@ -34,75 +34,73 @@ paths:
       responses:
         '200':
           description: >-
-            A list of configuration errors.
+            A list of configuration validation errors.
             An empty list indicates there are no errors.
           content:
             application/json:
               schema:
                 $ref: '../result/result.yaml#/components/schemas/Result.Errors'
-          
-  /control/port/link:
-    description: >-
-      Control port link API
-    post:
+    get:
       tags:
-        - Control
-      operationId: set_port_link
+        - Config
+      operationId: get_config
       description: >-
-        Sets the port link state on the traffic generator.
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '../control/control.yaml#/components/schemas/Control.PortLink'
+        Returns traffic generator configuration.
       responses:
-        '204':
+        '200':
           description: >-
-            No content
+            The most recent traffic generator configuration.
+            An empty configuration indicates that configuration does not exist.
+          content:
+            application/json:
+              schema:
+                $ref: '../config/config.yaml#/components/schemas/Config'
 
-  /control/port/capture:
+  /control/action:
     description: >-
-      Control port capture API
+      Control action API
     post:
       tags:
         - Control
-      operationId: set_port_capture
+      operationId: set_control_action
       description: >-
-        Starts a port capture on the traffic generator.
-        
-        Use the /results/capture API to stop a port capture and get the captured
-        results.
+        Applies a list of control actions on intended traffic generator resources
       requestBody:
         required: true
         content:
           application/json:
             schema:
-              $ref: '../control/control.yaml#/components/schemas/Control.PortCapture'
+              $ref: '../control/control.yaml#/components/schemas/Control.Action'
       responses:
-        '204':
+        '202':
           description: >-
-            No content
+            The request for intended control actions have been accepted and is in progress.
+            The response body will contain status url which will provide subsequent progress status.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status_url:
+                    type: string
 
-  /control/flow/transmit:
+  /control/status:
     description: >-
-      Control flow transmit API
-    post:
+      Control status API
+    get:
       tags:
         - Control
-      operationId: set_flow_transmit
+      operationId: get_control_status
       description: >-
-        Sets the flow transmit state on the traffic generator.
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '../control/control.yaml#/components/schemas/Control.FlowTransmit'
+        Returns status for all supported control actions.
       responses:
-        '204':
+        '200':
           description: >-
-            No content
+            A key-value pair of each supported action and its corresponding status.
+          content:
+            application/json:
+              schema:
+                $ref: '../control/control.yaml#/components/schemas/Control.Status'
 
   /results/capabilities:
     description: >-
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -11,6 +11,7 @@ components:
         A container for all models that are part of the configuration.
       type: object
       properties:
+        name: '../common/common.yaml#/components/x-inline/UniqueName'
         ports:
           description: >-
             The ports that will be configured on the traffic generator.
--- a/control/control.yaml
+++ b/control/control.yaml
@@ -8,58 +8,150 @@ info:
 
 components:
   schemas:
-    Control.PortLink:
+    Control.Action:
       description: >-
-        Control port link state
-      type: object
-      properties:
-        port_names:
-          description: >-
-            The names of port objects to.
-            An empty list will control all port objects.
-          type: array
-          items:
-            type: string
-            x-constraint:
-            - '/components/schemas/Port/properties/name'
-        state:
-          description: >-
-            The link state.
-          type: string
-          enum: [up, down]
-
-    Control.FlowTransmit:
-      description: >-
-        Control flow transmit state
+        One or more control actions to be performed against intended resources.
       type: object
       properties:
         flow_names:
           description: >-
-            The names of flow objects to control.
-            An empty list will control all flow objects.
+            The names of flows to apply control actions on.
+            An empty list will apply control actions on all flows.
           type: array
           items:
             type: string
             x-constraint:
             - '/components/schemas/Flow/properties/name'
-        state:
-          description: >-
-            The transmit state.
-          type: string
-          enum: [start, stop, pause]
-
-    Control.PortCapture:
-      description: >-
-        Control port capture state
-      type: object
-      properties:
         port_names:
           description: >-
-            The name of a port to start capturing packets on.
-            Allows for capture to be started on multiple ports.
+            The names of ports to apply control actions on.
+            An empty list will apply control actions on all ports.
           type: array
           items:
             type: string
             x-constraint:
             - '/components/schemas/Port/properties/name'
+        actions:
+          description: >-
+            List of control actions. The actions prefixed with `o_` are granular actions.
+          type: array
+          items:
+            type: string
+            enum:
+              - start
+              - stop
+              - pause
+              - resume
+              - update
+              - o_connect
+              - o_disconnect
+              - o_update_layer1
+              - o_start_transmit
+              - o_stop_transmit
+              - o_start_capture
+              - o_stop_capture
+              - o_clear_stats
+              - o_start_devices
+              - o_stop_devices
 
+    Control.Status:
+      description: >-
+        Status of all supported control actions. Whenever a control action is
+        requested, its status is reset to `none` and updated to `done` or `failed`
+        only when the underlying operations associated with it succeeds or fails
+        respectively. The details of failure for a given action can be found in
+        list of errors.
+      type: object
+      properties:
+        start:
+          description: >-
+            Indicates whether traffic run has been successfully started.
+          type: string
+          enum: [none, done, failed]
+          default: none
+        stop:
+          description: >-
+            Indicates whether traffic run has been successfully stopped.
+          type: string
+          enum: [none, done, failed]
+          default: none
+        pause:
+          description: >-
+            Indicates whether traffic run has been successfully paused.
+          type: string
+          enum: [none, done, failed]
+          default: none
+        resume:
+          description: >-
+            Indicates whether traffic run has been successfully resumed.
+          type: string
+          enum: [none, done, failed]
+          default: none
+        update:
+          description: >-
+            Indicates whether traffic config has been successfully updated.
+          type: string
+          enum: [none, done, failed]
+          default: none
+        o_connect:
+          description: >-
+            Indicates whether connection to intended traffic endpoints have been successfully established.
+          type: string
+          enum: [none, done, failed]
+          default: none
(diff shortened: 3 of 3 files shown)


ajbalogh:
other than the comment 'start traffic run' how does a user differentiate between starting traffic, starting capture, sending a ping on a specific number of devices etc.

ajbalogh:
Added GET /control/status: For checking statuses of previously invoked actions

What happens in a multi user env - if user 1 submits an action and user 2 submits an action - the get /control/status as modeled only returns the last action status?

ashutshkumr:
## Equivalent example for async request / response where progress url is returned with status code 202

Please read "NOTE" mentioned in the comments since few things are still not clear to me.

```python
from athena import *


def check_progress(progress_url):
    '''
    returns `in_progress`, `done` or `failed` after checking progress URL
    '''
    api = ApiClient()
    # empty header should be replaced with credential info
    data, status, headers = api.call_api(progress_url, 'GET', header_params={})
    if status == 200:
        # NOTE: the model for progress data is not defined in spec, so I'm assuming
        # it will contain `state` field indicating progress
        if data.state == 'IN_PROGRESS':
            return 'in_progress'
        elif data.state == 'COMPLETED':
            return 'done'
        elif data.state == 'FAILED':
            return 'failed'
    else:
        print('HTTP failure: %d' % status)
        return 'failed'


config_api = ConfigApi()
control_api = ControlApi()
results_api = ResultsApi()
cfg = Config(name='test', ports=[Port(name='p1')], flows=[Flow(name='f1')])

# POST configuration
data, status, headers = config_api.set_config_with_http_info(cfg)

# wait for configuration to be applied
if status == 202:
    while True:
        p = check_progress(response.progress_url)
        if p.start == 'done':
            break
        if p.start == 'failed':
            print("Config apply failed")
            break
elif status == 204:
    print('Done')
elif status == 400:
    print('Config apply failed')

# start traffic run
control_api.set_flow_transmit(ControlFlowTransmit(state='start'))

# wait for traffic run to be started
while True:
    response = results_api.get_flow_results(
        ResultFlowRequest(flow_names=[f.name for f in cfg.flows])
    )
    
    done = True

    # ideally we'll do this only once and also check if index is >= 0
    index = response.columns.index('state')

    # NOTE: how do we determine if starting a flow failed ?
    for row in response.rows:
        if row[index] != 'started':
            done = False
            break
```

ashutshkumr:
> other than the comment 'start traffic run' how does a user differentiate between starting traffic, starting capture, sending a ping on a specific number of devices etc.

This is why I have distinguished between 'composite' and 'granular' actions. e.g. `o_start_transmit` and `o_start_capture` will do exactly that, whereas `start` will do whatever it needs to do to start traffic run (which includes, stopping existing traffic run, clearing stats and even starting capture on ports that are configured for capture).

ashutshkumr:
> Added GET /control/status: For checking statuses of previously invoked actions
> 
> What happens in a multi user env - if user 1 submits an action and user 2 submits an action - the get /control/status as modeled only returns the last action status?

@ankur-sheth @winstonliu1111 

We admittedly haven't discussed how we're going to handle multi-user scenario throughout athena, but here's what I had as an initial proposal.

- User logs in using `Basic Authentication` (this is supported by OpenAPI)
- Every time a user sends any other request (including logout), credentials are sent as part of the header. So, every request has a context associated with its user.
- Thus, `/control/action` from user A will be applicable only to `/config` sent by user A. And user A sees `/control/status` which is applicable only to `/control/action` sent by user A.

ashutshkumr:
## Equivalent example for async request / response where progress ID is returned (grpc / protobuf)

This is very similar to HTTP 202 with progress_url example. Except, we use progress_id.

```python

import grpc
from athena import *

channel = grpc.secure_channel()
config_api = ConfigServiceStub(channel)
control_api = ControlServiceStub(channel)
results_api = ResultsServiceStub(channel)


def check_progress(progress_id):
    '''
    returns `in_progress`, `done` or `failed` after checking progress URL
    '''
    req = ProgressRequest(progress_id=progress_id)
    state = config_api.GetProgress(req).state

    # NOTE: the model for progress data is not defined in spec, so I'm assuming
    # it will contain `state` field indicating progress
    if state == 'IN_PROGRESS':
        return 'in_progress'
    elif state == 'COMPLETED':
        return 'done'
    elif state == 'FAILED':
        return 'failed'

cfg = ConfigRequest(name='test', ports=[Port(name='p1')], flows=[Flow(name='f1')])
# equivalent to POST /config
response = config_api.SetConfig(cfg)

# wait for configuration to be applied
while True:
    p = check_progress(response.progress_id)
    if p.start == 'done':
        break
    if p.start == 'failed':
        print("Config apply failed")
        break

# start traffic run
control_api.SetFlowTransmit(ControlFlowTransmitRequest(state='start'))

# wait for traffic run to be started
while True:
    response = results_api.GetFlowResults(
        ResultFlowRequest(flow_names=[f.name for f in cfg.flows])
    )
    
    done = True

    # ideally we'll do this only once and also check if index is >= 0
    index = response.columns.index('state')

    # NOTE: how do we determine if starting a flow failed ?
    for row in response.rows:
        if row[index] != 'started':
            done = False
            break

```

ankur-sheth:
> > other than the comment 'start traffic run' how does a user differentiate between starting traffic, starting capture, sending a ping on a specific number of devices etc.
> 
> This is why I have distinguished between 'composite' and 'granular' actions. e.g. `o_start_transmit` and `o_start_capture` will do exactly that, whereas `start` will do whatever it needs to do to start traffic run (which includes, stopping existing traffic run, clearing stats and even starting capture on ports that are configured for capture).

Do we need the composite actions yet?  It might not be obvious to a casual reader that it might start traffic and capture at the same time.  I assume that it should be possible to do a POST /control/action with a list of granular actions?

ankur-sheth:
> > Added GET /control/status: For checking statuses of previously invoked actions
> > What happens in a multi user env - if user 1 submits an action and user 2 submits an action - the get /control/status as modeled only returns the last action status?
> 
> @ankur-sheth @winstonliu1111
> 
> We admittedly haven't discussed how we're going to handle multi-user scenario throughout athena, but here's what I had as an initial proposal.
> 
>     * User logs in using `Basic Authentication` (this is supported by OpenAPI)
> 
>     * Every time a user sends any other request (including logout), credentials are sent as part of the header. So, every request has a context associated with its user.
> 
>     * Thus, `/control/action` from user A will be applicable only to `/config` sent by user A. And user A sees `/control/status` which is applicable only to `/control/action` sent by user A.

Ok.  What about ownership of resources (like ports)?  Is that implied?  Let's say we have the following scenario:

1. Controller up and ready.  Traffic engine A is up and ready.
2. User1 logs in, sends config with Port(location="A") and rest of the configuration as desired.
3. User1 starts traffic on A, stops traffic on port A and does NOT log out.
4. After few hours, User2 logs in and tries to send config with Port(location="A").  Will his POST fail?

ankur-sheth:
> What's new ?
> 
>     * Rephrased purpose of POST /config: It will validate the Config and store it (and not actually apply it).
> 

What will actually apply the configuration?  When will the controller actually try to connect to a port?  And how will the user know that say a given port is not up?  Which request needs to be called?

winstonliu1111 (commented):
few comments -
1) i'm not sure about the "imperative" actions becoming a unique action/status. 
for example, instead of say currently where things are control/action and control/status, where we report on start/stop for example for a traffic engine, maybe it should be

control/desiredOperationalState where if you want to control say traffic engine operational state,
  trafficEngineOperationalState = [Stopped, Started, Paused]

and control/actualOperationalState may be say
  trafficEngineOperationalState = [Stopped, Stopping, Started, Starting, Paused, Pausing]

and there should always be a way to do get on /control/desiredOperationalState to know what the user intents are and a way to get the actual operational state.  Don't mix and match/overload the concepts with stats/metrics.

2) andy raised a good point about the multi user scenario.  makes me wonder, how do we solve that in the current implementation, it would have similar multi user concerns, no?

ashutshkumr:
> > > other than the comment 'start traffic run' how does a user differentiate between starting traffic, starting capture, sending a ping on a specific number of devices etc.
> > 
> > 
> > This is why I have distinguished between 'composite' and 'granular' actions. e.g. `o_start_transmit` and `o_start_capture` will do exactly that, whereas `start` will do whatever it needs to do to start traffic run (which includes, stopping existing traffic run, clearing stats and even starting capture on ports that are configured for capture).
> 
> Do we need the composite actions yet? It might not be obvious to a casual reader that it might start traffic and capture at the same time. I assume that it should be possible to do a POST /control/action with a list of granular actions?

It's true that a `sequence` of granular actions should be able to replicate the sequence of underlying operations executed when a composite action is invoked. But, not exposing composite actions results in following assumptions:
- Script writers are aware of the exact sequence of granular actions required to start traffic session
- Script writers are aware of the duration of gap required between each granular action
- Script writers are aware of the difference in sequence on tx and rx ports

e.g. IxTclHal makes that assumption too, but Scapy doesn't (it's not a fair comparison, but FYI). IxN lies somewhere in the middle.

```python
# scapy
sendp(Ether()/IP(dst="1.2.3.4",ttl=(1,4)), iface="eth1")

# IxTclHal
ixStopTransmit txPort
ixStopCapture rxPort
ixStopPgid rxPort
ixClearStats txPort rxPort
ixClearTimestamps txPort rxPort
ixStartPgid rxPort
ixStartCapture rxPort
ixStartTransmit txPort
ixCheckTransmitDone txPort
```

ashutshkumr:
> > > Added GET /control/status: For checking statuses of previously invoked actions
> > > What happens in a multi user env - if user 1 submits an action and user 2 submits an action - the get /control/status as modeled only returns the last action status?
> > 
> > 
> > @ankur-sheth @winstonliu1111
> > We admittedly haven't discussed how we're going to handle multi-user scenario throughout athena, but here's what I had as an initial proposal.
> > ```
> > * User logs in using `Basic Authentication` (this is supported by OpenAPI)
> > 
> > * Every time a user sends any other request (including logout), credentials are sent as part of the header. So, every request has a context associated with its user.
> > 
> > * Thus, `/control/action` from user A will be applicable only to `/config` sent by user A. And user A sees `/control/status` which is applicable only to `/control/action` sent by user A.
> > ```
> 
> Ok. What about ownership of resources (like ports)? Is that implied? Let's say we have the following scenario:
> 
> 1. Controller up and ready.  Traffic engine A is up and ready.
> 2. User1 logs in, sends config with Port(location="A") and rest of the configuration as desired.
> 3. User1 starts traffic on A, stops traffic on port A and does NOT log out.
> 4. After few hours, User2 logs in and tries to send config with Port(location="A").  Will his POST fail?

Yes, it's implied. Since apiserver knows request mentioned in 4. is coming from user2, and user1 is already using endpoint A, it will raise conflict unless one of following happens:
- user A logs out
- user A's session times out due to inactivity
- apiserver is restarted

NOTE: The model currently does not provide any constructs for managing sessions or ownerships. So this is the best apiserver can do.

ashutshkumr:
> > What's new ?
> > ```
> > * Rephrased purpose of POST /config: It will validate the Config and store it (and not actually apply it).
> > ```
> 
> What will actually apply the configuration? When will the controller actually try to connect to a port? And how will the user know that say a given port is not up? Which request needs to be called?

All that happens as part of `start` (composite action).
From our previous discussions, I thought we agreed on following workflow:
- User shall just push the configuration and start traffic run (which in turn should take care of all needed granular actions.)

Otherwise, we have granular actions `o_connect`, `o_disconnect`, `o_update` (and can add `o_apply`).

ashutshkumr:
> few comments -
> 
> 1. i'm not sure about the "imperative" actions becoming a unique action/status.
>    for example, instead of say currently where things are control/action and control/status, where we report on start/stop for example for a traffic engine, maybe it should be
> 
> control/desiredOperationalState where if you want to control say traffic engine operational state,
> trafficEngineOperationalState = [Stopped, Started, Paused]
> 
> and control/actualOperationalState may be say
> trafficEngineOperationalState = [Stopped, Stopping, Started, Starting, Paused, Pausing]
> 
> and there should always be a way to do get on /control/desiredOperationalState to know what the user intents are and a way to get the actual operational state. Don't mix and match/overload the concepts with stats/metrics.
> 
> 1. andy raised a good point about the multi user scenario.  makes me wonder, how do we solve that in the current implementation, it would have similar multi user concerns, no?

For multi-user scenario, please see https://github.com/open-traffic-generator/models/pull/12#issuecomment-701534573 and https://github.com/open-traffic-generator/models/pull/12#issuecomment-702145365

For desired vs actual state approach, do you mean following ?
```sh
# start traffic
POST /control/desired
{
  "traffic": "started"
}

# get traffic state - try 1
GET /control/actual
{
  "traffic": "starting"
}

# get traffic state - try 2
GET /control/actual
{
  "traffic": "started"
}
```

If so, I like this approach too (since this is model driven as well), but I'd like to accommodate following:
- Add a `failed` state to the mix to definitively imply failure in attaining desired state
- Add an error message to provide error details
- Identify other resources which may need state change, apart from `traffic`

About mixing stats with control states, I'm not disagreeing, but I think we're already doing that. e.g. /results/flow has columns reflecting flow state. Similarly port stats has a column reflecting link state.

ashutshkumr:
Closing this since there's another alternative that's WIP.

PR #24, opened 2020-10-20
Multi speed gbe proposal

ajbalogh:
It has been raised during the sonic code review that the one_hundred_gbe with speed property is confusing.
To reduce confusion (speed and unused properties depending on speed) this PR splits creates individual objects for 10/40/100 gbe.
This allows for adding 200/400 gbe etc.

### DIFF
--- a/layer1/layer1.yaml
+++ b/layer1/layer1.yaml
@@ -15,7 +15,7 @@ components:
         name: '../common/common.yaml#/components/x-inline/UniqueName'
         port_names:
           description: >- 
-            A list of unique names of a port objects that will share the 
+            A list of unique names of port objects that will share the
             choice settings. 
           type: array
           items:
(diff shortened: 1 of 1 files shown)


ajbalogh:
Please expedite your reviews as this has the potential to block the sonic acceptance of test cases.

ajbalogh:
Sample test case using the new model.

```python
    port1 = Port(name='port1', location='10.36.74.26;01;01')
    port2 = Port(name='port2', location='10.36.77.102;12;03')
    port3 = Port(name='port3', location='10.36.77.102;12;10')
    port4 = Port(name='port4', location='10.36.77.102;12;16')
    ethernet = Layer1(name='ethernet 10/100/1000 settings',
                      port_names=[port1.name],
                      choice=Ethernet(media='copper',
                                      speed='one_thousand_mbps',
                                      auto_negotiate=True))
    novus_hundred = Layer1(name='novus 100gbe settings',
                           port_names=[port2.name],
                           choice=OneHundredGbe(ieee_media_defaults=False,
                                                auto_negotiate=False,
                                                link_training=False,
                                                rs_fec=True))
    novus_forty = Layer1(name='novus 40gbe settings',
                         port_names=[port3.name],
                         choice=FortyGbe(ieee_media_defaults=False,
                                        auto_negotiate=False,
                                        link_training=False,
                                        rs_fec=True))
    novus_ten = Layer1(name='novus 10gbe settings',
                         port_names=[port4.name],
                         choice=TenGbe())
    config = Config(ports=[port1, port2, port3],
                    layer1=[ethernet, novus_hundred, novus_forty, novus_ten])
    api.set_state(State(ConfigState(config=config, state='set')))
```

abhijit-dhar:
This looks very good to me.  Making Ethernet layers as a choice of Layer1 parameter  is far better approach than setting -speed switch with different values.   Note in the sample script there should be port4 as well. 
```
config = Config(ports=[port1, port2, port3, port4],
                    layer1=[ethernet, novus_hundred, novus_forty, novus_ten])

```
Will there be any error message if this lists mismatch?

ashutshkumr:
Options for using FC FEC, RS FEC, KP4 FEC, Auto Negotiation, LT, etc may vary not only based on `speed` but also based on `cable` (fiber vs copper vs SGMII), `transceiver` (QSFPDD vs QSFP28) and `encoding` (NRZ vs PAM) type.

Hence, instead of having speed-specific containers, I would propose to have containers for `those specific` settings in `Layer1` itself. Speed should simply be an enum choice.

NOTE: I have also proposed below prefixing speed ENUMs with `speed` so we can use numeric values instead of keeping it all wordy.

e.g.

```python
c = Config(
    ports=[
        Port(name='p1'), Port(name='p2'), Port(name='p3'), Port(name='p4')
    ],
    layer1=[
        Layer1(
            name='p1-p2-sfp', port_names=['p1', 'p2'], speed='speed_100_mbps',
            media='fiber', auto_negotiate=True
        ),
        Layer2(
            name='p3-p4-cr4', port_names=['p3', 'p4'], speed='speed_100_gbps',
            media='copper', use_ieee_defaults=True
        )
    ]
)
```

ajbalogh:
@ashutshkumr  in your last comment you indicated specific containers for settings like fec but they aren't present in the sample - what would that end up looking like? Does that mean a user selects speed and chooses from settings that may or may not apply to a speed?

ashutshkumr:
> Does that mean a user selects speed and chooses from settings that may or may not apply to a speed?

Indeed.

> @ashutshkumr in your last comment you indicated specific containers for settings like fec but they aren't present in the sample - what would that end up looking like? 

I do not have any example off the top of my head. We'll need to compile all possible MVP L1 settings we are targeting to support. The only point I was trying to make was that those settings need not be dependent only on speed (and hence we shouldn't have speed-specific containers).

Just an e.g. I came up impromptu, *it's not a proposal*.
```python
# when ieee defaults is True, no need to provide any AN/RS-FEC settings
a = Layer1(
    name='100g-base-cr4', port_names=['p1', 'p2'], speed='speed_100_gbps',
    media='copper', use_ieee_defaults=True
)
# when ieee defaults is False, users can control AN and RS-FEC settings
a = Layer1(
    name='100g-base-cr4', port_names=['p1', 'p2'], speed='speed_100_gbps', media='copper',
    fc_fec=FCFec(enable=True, enable_stats=True),
    auto_negotiation=AutoNegotiation(enable=False, enable_link_training=False)
)
```

PR #26, opened 2020-10-23
Layer1 settings with speed as enum

ajbalogh:
This is an alternate model approach proposed by @ashutshkumr to the multi-speed-gbe-proposal.
The intent is to have speed as an enum at the top level of layer1 with specific containers for ethernet, gigabit_ethernet and virtual_ethernet settings, and flow control.

### DIFF
--- a/layer1/layer1.yaml
+++ b/layer1/layer1.yaml
@@ -10,33 +10,88 @@ components:
       description: >-
         A container for layer1 settings.
       type: object
-      required: [name, choice]
+      required: [name, port_names]
       properties:
         name: '../common/common.yaml#/components/x-inline/UniqueName'
         port_names:
           description: >- 
-            A list of unique names of a port objects that will share the 
+            A list of unique names of port objects that will share the
             choice settings. 
           type: array
           items:
             type: string
             x-constraint:
             - '/components/schemas/Port/properties/name'
-        choice:
+        speed:
           description: >-
-            The type of layer1 characteristics.
+            The speed of the port.
+            If the speed is not supported an error will be returned.
           type: string
-          enum: [one_hundred_gbe, ethernet, ethernet_virtual]
-        one_hundred_gbe:
-          $ref: '#/components/schemas/Layer1.OneHundredGbe'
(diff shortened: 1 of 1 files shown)


ashutshkumr (approved):
Ok from my end on changes related to speed enums (and addition of few more settings). But we may need to further discuss on possible changes that can be done to `ethernet`, `gigabit_ethernet` and `virtual_ethernet` containers as tracked in https://github.com/open-traffic-generator/models/pull/24

layer1/layer1.yaml
```diff
@@ -10,124 +10,112 @@ components:
...
+        ieee_media_defaults:
           description: >-
-            Enable/disable reed solomon forward error correction (RS FEC).
+            Set to true to override the auto_negotiate, link_training
+            and rs_fec settings for gigabit ethernet interfaces.
           type: boolean
-          default: false
-        speed:
+          default: true
+        auto_negotiate:
```
ashutshkumr, line 64: We don't need this anymore (since all AN-related settings should now be part of AutoNegotiation)

layer1/layer1.yaml
```diff
@@ -10,124 +10,112 @@ components:
...
           type: boolean
           default: true
-        advertise_one_thousand_mbps:
+        advertise_100_fd_mbps:
           description: >-
-            If auto_negotiate is true then this speed will be advertised.
+            If auto_negotiate is true and the interface supports this option
+            then this speed will be advertised.
           type: boolean
           default: true
```
ashutshkumr, line 90: Two queries:
- Shouldn't all `advertise_*` default to false ? i.e. doesn't make sense to have advertise_100_fd_mbps and advertise_100_hd_mbps set to true at the same time. They also are not needed when rs_fec=True (since it applies to gigabit eth).
- Is FD (full duplex) and HD (half duplex) documented somewhere ?

ashutshkumr (approved):
Please merge regardless of whether two minor pending concerns are addressed (they can be further discussed upon separately, if needed).

PR #37, opened 2020-12-10
Added a few missing fields to TCP flow header

### LINKED ISSUE #35: TCP Packet Header is missing the window parameter
The TCP packet header is missing the 16-bit window parameter.

ankur-sheth:
Added sequence number, acknowledgement number, data_offset, and window parameters to the TCP flow definition.  Closes #35.


ajbalogh (approved):

ashutshkumr (approved):
Would it make sense to shorten respective field names to `seq_num` and `ack_num` ?

Need to check with team on how we missed this (should've been raised before).

PR #51, opened 2021-01-27
add custom and packet captures

ajbalogh:
Refactor for a more agnostic capture specification.
The following is what it would be like to specify a capture configuration using the current proposal.

```python
def test_capture(api):
    """Test capture functionality

    Capture should allow for multiple filters.
    The capture.filters is a list of CaptureFilter objects which can be 
    CaptureCustom, CaptureEthernet, CaptureVlan and CaptureIpv4.
    A CaptureCustom object is used when packet header is not available.
    """
    config = api.config()

    # create a capture
    config.ports.port(name='port1').port(name='port2')
    capture = config.captures.capture()[0]
    capture.name = 'capture1'
    capture.port_names = [port.name for port in config.ports]

    # add 2 custom filters
    capture.filters.custom().custom()
    src_mac = capture.filters[0]  # type: snappi.CustomFilter
    src_mac.offset = 0
    src_mac.value = '000100000001'
    src_mac.mask = 'FFFFFFFFFFFF'
    dst_mac = capture.filters[1]  # type: snappi.CustomFilter
    dst_mac.offset = 0
    dst_mac.value = '000200000001'
    dst_mac.mask = 'FFFFFFFFFFFF'
    print(config)

    # add packet filters for inner and outer vlan
    capture.filters.clear()
    _, i_vlan, o_vlan = capture.filters.ethernet().vlan().vlan()
    i_vlan.id.value = '01'
    i_vlan.id.mask = 'FF'
    o_vlan.id.value = '03'
    o_vlan.id.mask = 'FF'
    print(config)

    # add packet filters for a custom packet after eth, ip
    capture.filters.clear()
    eth, _, custom = capture.filters.ethernet().ipv4().custom()
    eth.src.value = '000001000001'
    eth.src.mask = 'FFFFFFFFFFFF'
    custom.offset = 16
    custom.value = '010101'
    custom.mask = 'FFFFFF'
    print(config)
```

### DIFF
--- a/capture/capture.yaml
+++ b/capture/capture.yaml
@@ -12,7 +12,7 @@ components:
       description: >-
         Container for capture settings.
       type: object
-      required: [name, choice]
+      required: [name]
       properties:
         port_names:
           description: >-
@@ -22,24 +22,22 @@ components:
             type: string
             x-constraint:
             - '/components/schemas/Port/properties/name'
-        choice: 
+        custom_filters:
           description: >-
-            The type of filter.
-          type: string
-          enum: [basic, pcap]
-        basic:
-          description: >-
-            An array of basic filters.
-            The filters supported are source address, destination address and
-            custom.  
+            A list of custom filters
           type: array
           items:
-            $ref: '#/components/schemas/Capture.BasicFilter'
-        pcap:
+            $ref: '#/components/schemas/Capture.Custom'
+        packet_filter:
           description: >-
-            The content of this property must be of pcap filter syntax.
-            https://www.tcpdump.org/manpages/pcap-filter.7.html
-          type: string
+            A list of packet filters to apply to the capturing ports.
+            If no packet filters are specified then all packets will be captured.
+            A capture can have multiple filters. The number of filters supported
+            is determined by the implementation which can be retrieved using
+            the capabilities API.
+          type: array
+          items:
+            $ref: '#/components/schemas/Capture.Filter'
         enable:
           description: >-
             Enable capture on the port.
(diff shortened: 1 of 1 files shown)


capture/capture.yaml
```diff
@@ -57,64 +55,72 @@ components:
...
-          description: >-
-            The mask to be applied to the filter.
           type: string
-        offset:
-          description: >-
-            The offset in the packet to filter at.
-          type: integer
+          format: hex
+
+    Capture.Ethernet:
```
ankur-sheth, line 98: Do we want to add all these protocol headers in this file itself?  Will become a fairly long file as we add the other headers.
Also, any particular reason why you've only added Ethernet, VLAN and IPv4?

ankur-sheth (approved):
See one comment further down.

capture/capture.yaml
ajbalogh: The protocol headers can be moved to separate files once the proposal is accepted.  I've only added ethernet, vlan, ipv4 and custom to illustrate how the capture.filters creation is similar to flow.packet.  Adding additional protocols is a trivial exercise, and an extension like x-replace can be introduced so we can reuse the flow.packet headers and just replace the pattern ref with the capture field ref.

ashutshkumr:
```python
a, b = capture.filters.custom().custom()
```

Does this filter `a && b` or `a || b` ?

ankur-sheth (approved):

ashutshkumr:
This looks really neat and simple. Although, since we're refactoring capture, I would suggest considering inclusion of following:
- packet count
- capture size of packet

Equivalent to `tcpdump -c 2 -s 100` (2 packets each of slice size 100B). Whether or not we need first 2 or last 2 packets is guided by `overwrite`.

P.S. I was also looking for a way to use boolean operator other than `&&`, but I think this model is in shape so that it can be expanded further later when required.

ashutshkumr (approved):

PR #56, opened 2021-01-28
Add Arp packet header

indraniBh:
Added support for ARP header

### DIFF
--- a/flow/packet-headers/arp.yaml
+++ b/flow/packet-headers/arp.yaml
@@ -1,0 +1,58 @@
+openapi: 3.1.0
+
+info:
+  title: Arp packet header model
+  version: ^0.0.0
+
+components:
+  schemas:
+    Flow.Arp:
+      description: Arp packet header
+      type: object
+      properties:
+        hardware_type:
+          description: >-
+            Network link protocol type. 
+            Max length is 2 bytes.
+          $ref: './patterns.yaml#/components/schemas/Flow.Pattern'
+        protocol_type:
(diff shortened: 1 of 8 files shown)


flow/packet-headers/arp.yaml
```diff
@@ -0,0 +1,58 @@
+openapi: 3.1.0
+
+info:
+  title: Arp packet header model
+  version: ^0.0.0
+
+components:
+  schemas:
+    Flow.Arp:
+      description: Arp packet header
```
ankur-sheth, line 10: Change to ARP instead of Arp in the description.

ankur-sheth (changes requested):

flow/packet-headers/arp.yaml
ashutshkumr: This (all caps in description) will require change for other protocol descriptions as well. e.g. https://github.com/open-traffic-generator/models/blob/master/flow/packet-headers/udp.yaml#L10

(I agree with the change)

ankur-sheth: I didn't look closely at the other headers.  Glanced at a couple and they seemed ok.  Regardless, it should be ARP in the description.

indraniBh: Updated description.

ajbalogh (approved):

ankur-sheth (approved):

ankur-sheth:
Merge ARP packet header model.

PR #70, opened 2021-02-10
Add burst_count in burst in duration

indraniBh:
burst_count provides number of burst per stream. Default is 0 which means continuous burst.

### DIFF
--- a/flow/duration.yaml
+++ b/flow/duration.yaml
@@ -101,13 +101,15 @@ components:
 
     Flow.Burst:
       description: >-
-        A continuous burst of packets that will not automatically stop.
+        Transmits continuous or fixed burst of packets. 
+        For continuous burst of packets, it will not automatically stop.
+        For fixed burst of packets, it will stop after transmitting fixed number of bursts.      
       type: object      
       properties:
-        burst_count:
+        bursts:
           description: >-
-            The number of burst transmitted per stream.
-            A value 0 means it is continuous burst.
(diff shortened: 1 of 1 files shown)


flow/duration.yaml
```diff
@@ -104,6 +104,12 @@ components:
         A continuous burst of packets that will not automatically stop.
       type: object      
       properties:
+        burst_count:
+          description: >-
+            The number of burst transmitted per stream.
```
ashutshkumr, line 109: - please rephrase description to remove mention of stream. e.g. "The number of packet bursts transmitted per flow. A value of 0 implies continuous burst of packets."
- we should call it `bursts` (instead of burst_count) to be inline with `packets`

flow/duration.yaml
```diff
@@ -104,6 +104,12 @@ components:
         A continuous burst of packets that will not automatically stop.
```
ashutshkumr, line 104: please update this descriptions since this container will now support both continuous and fixed bursts.

ashutshkumr (changes requested):
needs changes

flow/duration.yaml
indraniBh: Done.

flow/duration.yaml
indraniBh: Done.

ajbalogh (approved):

ashutshkumr (approved):

PR #74, opened 2021-02-15
Dev latency

indraniBh:
Added latency stats

### DIFF
--- a/result/flow.yaml
+++ b/result/flow.yaml
@@ -39,6 +39,9 @@ components:
               - bytes_rx
               - frames_tx_rate
               - frames_rx_rate
+              - min_latency
+              - max_latency
+              - avg_latency
               - loss
         metric_group_names:
           description: >-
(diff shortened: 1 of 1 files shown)


result/flow.yaml
```diff
@@ -113,6 +116,18 @@ components:
...
+          type: integer
+        max_latency:
+          description: >-
+            The maximum latency of valid frames received
+          type: integer
+        avg_latency:
+          description: >-
+            The avarage latency of valid frames received
+          type: number
         loss:
```
ashutshkumr, line 131: We need to specify unit of latency measurements (ns).
@ajbalogh should we include this in attr name `min_latency_ns` ? or just mention the units in description ?

ashutshkumr:
@ajbalogh @winstonliu1111 @winstonliu1111 should we add first and last timestamps as well to flow stats ? usecase is to calculate / validate total duration of each flow.

ashutshkumr (changes requested):
needs changes

result/flow.yaml
ajbalogh: Including units would keep it consistent with bytes_tx, frames_tx etc.

ajbalogh (changes requested):

winstonliu1111:
> @ajbalogh @winstonliu1111 @winstonliu1111 should we add first and last timestamps as well to flow stats ? usecase is to calculate / validate total duration of each flow.

I think we should wait and see how we deal with events and analysis in general as opposed to bake this into a stats path.  That is the next stop we need to model and resolve after we wrap first phase of this srte discussions.

result/flow.yaml
indraniBh: Done

ashutshkumr (approved):

PR #92, opened 2021-03-22
plugins sample

Rangababu-R:
@ajbalogh
This PR is just to get the gist of plugins way of implementation.
moved flow metrics from the base package and made it as plugin.
currently to identify plugin prepended flow metrics objects with string 'Plugin.'
ex: 'Plugin.Flow.Metrics.Request'
once if this approach is approved all the plugins can be moved to separate node instead of prepending the string.
Also modified snappigenerator.py to support plugin installations. Info will be available in snappi PR.

### DIFF
--- a/artifacts/openapi.yaml
+++ b/artifacts/openapi.yaml
@@ -2929,7 +2929,7 @@ components:
         port:
           $ref: '#/components/schemas/Port.Metrics.Request'
         flow:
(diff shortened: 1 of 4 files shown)


ajbalogh:
This is premature as we are still considering the benefits of a plugin/feature vs monolithic API.

PR #95, opened 2021-03-31
Event Metrics and Advanced Metrics API

ajbalogh:
What: An API proposal for event metrics and advanced metrics. 

[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/event-api/artifacts/openapi.yaml)

Sample Convergence Test Case
```
import pytest


@pytest.fixture
def primary_routes_name():
    return 'Rx Primary Routes'


@pytest.fixture
def secondary_routes_name():
    return 'Rx Secondary Routes'


@pytest.fixture
def convergence_config(api, primary_routes_name, secondary_routes_name):
    """A skeleton configuration suitable for testing convergence.
    Sets up TX routes and identical RX primary/secondary routes.
    """
    config = api.config()
    p_tx, p_rx_pri, p_rx_sec = config.ports \
        .port(name='Tx Port') \
        .port(name='Primary Rx Port') \
        .port(name='Secondary Rx Port')
    d_tx, d_rx_pri, d_rx_sec = config.devices \
        .device(name='Tx Device', container_name=p_tx.name) \
        .device(name='Primary Rx Device', container_name=p_rx_pri.name) \
        .device(name='Secondary Rx Device', container_name=p_rx_sec.name)
    tx_rr, = d_tx.ethernet.ipv4.bgpv4.bgpv4_routes.bgpv4route(name='Tx Routes')
    rx_prr, = d_rx_pri.ethernet.ipv4.bgpv4.bgpv4_routes.bgpv4route(name=primary_routes_name)
    rx_srr, = d_rx_sec.ethernet.ipv4.bgpv4.bgpv4_routes.bgpv4route(name=secondary_routes_name)
    flw, = config.flows.flow(name='Tx Ipv4 Device -> Rx Primary/Secondary Routes')
    flw.tx_rx.device.mode = flw.tx_rx.device.ONE_TO_ONE
    flw.tx_rx.device.tx_names = [tx_rr.name]
    flw.tx_rx.device.rx_names = [rx_prr.name, rx_srr.name]
    return config


def test_data_plane_convergence(api, convergence_config, primary_routes_name, secondary_routes_name):
    """Test data plane convergence using advanced metrics

    Configure convergence setup
    Configure advanced convergence metrics
    Create a convergence event
    Retrieve convergence metrics 
    """
    # set the traffic generator configuration
    api.set_config(convergence_config)

    # build the configuration for advanced convergence metrics
    adv_metrics = api.advancedmetrics_config()
    adv_metrics.convergence.enable = True
    adv_metrics.convergence.rx_rate_threshold = 90
    api.set_advanced_metrics(adv_metrics)

    # withdraw primary routes using the primary_routes_name
    route_state = api.route_state()
    route_state.state = route_state.WITHDRAW
    route_state.names = [primary_routes_name]
    api.set_route_state(route_state)

    # get advanced data plane convergence metrics
    adv_metrics_request = api.advancedmetrics_request()
    adv_metrics_request.convergence.flow_names = [
        convergence_config.flows[0].name
    ]
    adv_metrics_request.convergence.event_sources = [
        secondary_routes_name
    ]
    while True:
        adv_metrics = api.get_advanced_metrics(adv_metrics_request)
        if all([m.data_plane_convergence_ns is not None for m in adv_metrics.convergence_metrics]):
            break

    # output the convergence metrics
    print(adv_metrics.convergence_metrics)

    # fail the test if data plane convergence takes longer than 500ms
    assert(all([m.data_plane_convergence_ns < 500000000 for m in adv_metrics.convergence_metrics]))


if __name__ == '__main__':
    pytest.main(['-vv', '-s', __file__])
```


ashutshkumr:
Will it be feasible to facilitate a way to fetch tx/rx stats (frame count, loss, etc.) and latency stats for a given flow at same instance of time ?

Maybe we need to understand how important this use case is, but currently this is not achievable.

advanced/config.yaml
```diff
@@ -0,0 +1,90 @@
...
+  schemas:
+    AdvancedMetrics.Config:
+      description: |-
+        The container for advanced metrics configuration.
+      type: object
+      required: [choice]
+      properties:
+        choice:
+          type: string
+          enum: [events, convergence, latency]
```
ashutshkumr, line 11: Shouldn't we get rid of choice here ? since e.g. I should be able to enable both events and convergence metrics in one go.

advanced/config.yaml
```diff
@@ -0,0 +1,90 @@
...
+        events:
+          $ref: '#/components/schemas/AdvancedMetrics.Config.Events'
+        convergence:
+          $ref: '#/components/schemas/AdvancedMetrics.Config.Convergence'
+        latency:
+          $ref: '#/components/schemas/AdvancedMetrics.Config.Latency'
+
+    AdvancedMetrics.Config.Events:
+      type: object
+      properties:
```
ashutshkumr, line 21: I feel we should provide options to selectively enable certain kinds of events. This would mean that every event type (or a set of similar event types) will have a container with at least one attr called `enable`.

We can also expose `all` to allow enabling all supported events.

ashutshkumr: To expand more on this with examples.

```python
mcfg = api.advancedmetrics_config()
# will result in two possible events, `flow_rx_rate_above_threshold`, `flow_rx_rate_below_threshold`
mcfg.events.flow_rx_rate.enable = True
mcfg.events.flow_rx_rate.threshold_value = 95
# will result in two possible events, `port_link_up`, `port_link_down`
mcfg.events.port_link.enable = True
```

advanced/metrics.yaml
```diff
@@ -0,0 +1,187 @@
...
+          enum:
+          - route_advertise
+          - route_withdraw
+          example: 'route_withdraw'
+        event_begin_ns:
+          description: |-
+            The timestamp of when the event began.
+          type: number
+          format: double
+        event_end_ns:
```
ashutshkumr, line 152: What does `event_end_ns` imply ? e.g. if somebody triggers `route_withdraw`, `event_begin_ns` will consist of the timestamp when it was triggered. But what would `event_end_ns` contain ?

advanced/metrics.yaml
```diff
@@ -0,0 +1,187 @@
...
+        latency_metrics:
+          type: array
+          items:
+            $ref: '#/components/schemas/AdvancedMetrics.Latency'
+        convergence_metrics:
+          type: array
+          items:
+            $ref: '#/components/schemas/AdvancedMetrics.Convergence'
+
+    AdvancedMetrics.Latency:
```
ashutshkumr, line 92: we should include `first_timestamp` and `last_timestamp` here

ashutshkumr:
There are multiple places we are using `*_timestamp` attributes. It would be nice if we could have one common description of what is the format of timestamp values. (unix timstamps ?)

advanced/config.yaml
```diff
@@ -0,0 +1,90 @@
...
+          maximum: 100
+          format: float
+          default: 95
+
+    AdvancedMetrics.Config.Convergence:
+      description: |-
+        The optional container for convergence configuration options.
+      type: object
+      properties:
+        enable:
```
ashutshkumr, line 43: Should enabling convergence automatically enable events as well (and allow fetching event metrics) ?

ANISH-GOTTAPU:
@ashutshkumr : sample test as discussed is updated here https://github.com/open-traffic-generator/snappi-tests/pull/15/files

advanced/metrics.yaml
```diff
@@ -0,0 +1,187 @@
...
+      type: object
+      properties:
+        flow_name:
+          description: |-
+            The name of the flow that is affected by the trigger event.
+          type: string
+          x-constraint:
+          - '/components/schemas/Flow/properties/name'
+          example: 'Tx Port -> Rx Port'
+        event_source_name:
```
ashutshkumr, line 132: Do we need to have event-related information here ?
- There are actually two events that we're internally relying on (above/below threshold and advertise/withdraw routes)
- These events should be available in event metrics

If we need to keep it then I suggest changing the descriptions to remove `BGP` since it is applicable to other protocols too.

ashutshkumr:
> @ashutshkumr : sample test as discussed is updated here https://github.com/open-traffic-generator/snappi-tests/pull/15/files

@ankur-sheth @ajbalogh @winstonliu1111

advanced/metrics.yaml
```diff
@@ -0,0 +1,187 @@
...
+          type: string
+          enum: [event, latency, convergence]
+        event:
+          $ref: '#/components/schemas/AdvancedMetrics.Event.Request'
+        latency:
+          $ref: '#/components/schemas/AdvancedMetrics.Latency.Request'
+        convergence:
+          $ref: '#/components/schemas/AdvancedMetrics.Convergence.Request'
+
+    AdvancedMetrics.Event.Request:
```
winstonliu1111, line 16: just want to understand the semantics of this.  I can see how metrics can be pulled and be consumed instantaneously (either as single shot or the consumer decides to shove that into a database after polling)

but does this pattern really apply to events?  events are usually fired and forget by the publisher.  Are we implying each traffic engine need to remember the last event triggered?  It doesn't feel like the right pattern for events.

advanced/metrics.yaml
winstonliu1111: if the user is interested in data_plane_convergence, should they just ask for 
AdvancedMetrics.DataplaneConvergence.time

and AdvancedMetrics.DataplaneConvergence.details for the detailed events/timestamps that are used to arrive at the "time"?

CPDP will then be
AdancedMetrics.ControlplaneDataplaneConvergence ?  

this will probably make the known "events" being used to arrive at the time be more specific to the specific methodology of interest?

api/advanced-metrics.yaml
```diff
@@ -0,0 +1,70 @@
...
+            Response from traffic generator for chosen metrics
+          content:
+            application/json:
+              schema:
+                $ref: '../advanced/metrics.yaml#/components/schemas/AdvancedMetrics.Response'
+
+
+  /event/subscribe:
+    post:
+      tags: ['Advanced Metrics']
```
winstonliu1111, line 46: see my comments earlier about common patterns related to events and generally events != metrics...something to think about.

api/advanced-metrics.yaml
```diff
@@ -0,0 +1,70 @@
...
+          application/json:
+            schema:
+              $ref: '../advanced/config.yaml#/components/schemas/AdvancedMetrics.Config'
+      responses:
+        '200':
+          $ref: '../result/request.yaml#/components/responses/Success'
+        '409':
+          $ref: '../result/request.yaml#/components/responses/Conflict'
+        '590':
+          $ref: '../result/request.yaml#/components/responses/Fail'
```
winstonliu1111, line 20: i think we mentioned at some point, one of these response codes will be the means an implementation detailed back to a client if some capabilities in the config is not supported.

is there any guidance on the format/response code implementation should adhere to?

winstonliu1111 (commented):
it's probably worth for us to get together in a meeting to chat about some of the points.

i have an uneasiness about semantics of events being grouped under metrics, how we expect events to be consumed (typically events are streaming/pushed, not pulled)

ashutshkumr:
@ByReaL @selldinesh did you get a chance to review the data model ? These changes affect how we'll end up writing community test cases for CP/DP.

api/advanced-events.yaml
```diff
@@ -0,0 +1,36 @@
+paths:
+  /event/subscribe:
```
winstonliu1111, line 2: Can we get a rough idea what this endpoint looks like in the generated protobuf definition?

I'm hoping this+callback maps more stragihtforward to native grpc streaming constructs?

ajbalogh: The POST /event/subscribe API call will translate into an rpc called subscribe_event (operationId) with the request payload as streaming input and the callback request payload as streaming output. 
```
rpc subscribe_event(stream Event.Subscription) returns (stream Event)
```

winstonliu1111: Sounds good.  Is this reflected in the generated artifact for org.proto?  The last I checked, it's returning details as opposed to stream Event.  Is there change to the openapi->proto generator required?

ajbalogh: Its not reflected at this time, once agreement is reached a change to the patch process will need to be made.

ajbalogh:
removing in favor of pr #108

PR #100, opened 2021-04-13
changing type for kbps/mbps/gbps from integer to number

biplamal:


flow/rate.yaml
```diff
@@ -26,21 +26,26 @@ components:
         bps:
           description: >-
             Bits per second.
+            default is maximum supported speed.
```
ajbalogh, line 29: Start sentences with upper case.

ajbalogh (changes requested):

ajbalogh (approved):

PR #105, opened 2021-04-14
refactor inter burst gap

ajbalogh:
Why: The current structure of inter burst gap doesn't follow the choice model. The choice model allows for different defaults per choice while the original model does not. The choice model also removes the need to set units and a value.

[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/inter_burst_gap_refactor/artifacts/openapi.yaml)

```
# the refactor allows for this snappi notation
flw.duration.burst.inter_burst_gap.microseconds = 1.5
```

### DIFF
--- a/artifacts/openapi.yaml
+++ b/artifacts/openapi.yaml
@@ -2744,18 +2744,38 @@ components:
           type: integer
           default: 12
         delay:
-          description: |-
-            The delay before starting transmission of packets.
-          type: integer
-          default: 0
-        delay_unit:
-          description: |-
-            The delay expressed as a number of this value.
+          $ref: '#/components/schemas/Flow.Delay'
+    Flow.Delay:
+      description: |-
+        The optional container to specify the delay before starting  transmission of packets.
+      type: object
+      required:
+      - choice
+      properties:
+        choice:
           type: string
           enum:
           - bytes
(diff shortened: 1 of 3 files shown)


ashutshkumr (approved):
I like this change.
Although upon review, I also found that `delay` and `delay_unit` suffer from the same problem. Also, it's missing from `burst` container (not sure if it's by design).

ajbalogh:
@ashutshkumr i agree with you that delay and delay_unit have the same problem, I will make a change to address that and ask you for one more review - thanks for looking at this more closely.

artifacts/openapi.yaml
```diff
@@ -2740,18 +2744,38 @@ components:
...
+        choice:
           type: string
           enum:
           - bytes
           - nanoseconds
-          default: bytes
+          - microseconds
+        bytes:
+          description: |-
+            The delay before starting transmission of packets. A value of 0 indicates no gap between bursts.
```
ashutshkumr, line 2763: Need to change description (since current one refers to burst)

ashutshkumr (approved):
please merge after the description fix. rest looks good to me.

PR #113, opened 2021-05-10
Default field values update

shramroy:


device/ethernet.yaml
```diff
@@ -19,7 +19,7 @@ components:
           type: integer
           minimum: 0
           maximum: 65535
-          default: 1500
+          default: 1518
```
ajbalogh, line 22: is this the commonly accepted default?

flow/flow.yaml
```diff
@@ -25,6 +25,7 @@ components:
             The header is a list of traffic protocol headers.
             The order of traffic protocol headers assigned to the list is the
             order they will appear on the wire.
+            If no packet header is provided Ethernet will be set as default header.
```
ajbalogh, line 28: Disagree with this, the packet list is what is configured, there is also no notation in openapi to explicitly define a default item for the list.

layer1/layer1.yaml
```diff
@@ -39,7 +39,7 @@ components:
             - speed_100_gbps
             - speed_200_gbps
             - speed_400_gbps
-          default: speed_10_gbps
+          default: speed_1_gbps
```
ajbalogh, line 42: Disagree as it was changed to be 10 from 1.

layer1/layer1.yaml
```diff
@@ -52,12 +52,12 @@ components:
...
             Enable promiscuous mode if supported.
           type: boolean
-          default: false
+          default: true
         mtu:
           description: >-
             Set the maximum transmission unit size if supported.
           type: integer
-          default: 1500
+          default: 1518
```
ajbalogh, line 60: This doesn't appear to be the common default.

ajbalogh (changes requested):

flow/flow.yaml
ashutshkumr: @ajbalogh @ankur-sheth won't it be more user-friendly if we configure a default ethernet header under the hood when no packet headers are provided ? not being able to specify default for lists is a limitation in openapi, hence we chose to mention it in description.

layer1/layer1.yaml
shramroy: Changed

layer1/layer1.yaml
shramroy: Changed

device/ethernet.yaml
shramroy: changed

flow/flow.yaml
ajbalogh: Reviewed the openapi spec and you can specify a default in this case by setting the minimum array length to 1.  
Be aware that every packet header that gets added now defaults to ethernet instead of custom.  

```yaml
in flow.yaml add minItems to the packet property

        packet:
          description: >-
            The header is a list of traffic protocol headers.
            The order of traffic protocol headers assigned to the list is the
            order they will appear on the wire.
          type: array
          minItems: 1
          items:
            $ref: './packet-headers/header.yaml#/components/schemas/Flow.Header'

in header.yaml change the choice property default to ethernet

    Flow.Header:
      description: >-
        Container for all traffic packet headers
      type: object
      required: [choice]
      properties:
        choice:
          type: string
          enum: [custom, ethernet, vlan, vxlan, ipv4, ipv6, pfcpause, ethernetpause, tcp, udp, gre, gtpv1, gtpv2, arp, icmp, icmpv6, ppp, igmpv1]
          default: ethernet

```

ajbalogh: Everything else lgtm but I will hold off on my review until this conversation is resolved

ajbalogh (commented):
LGTM except for default ethernet packet header discussion

flow/flow.yaml
shramroy: I have updated the changes according.

ashutshkumr: @shramroy could you please check and answer expected behavior here (I'm trying to understand how minItems works) ?
- user specifies flow without specifying packets
- user specifies flow with packets where list is empty

In both the cases I'm assuming we'll end up with packets containing list of one header, which defaults to ethernet (at least controller needs to ensure it).
If so, the change LGTM.

ajbalogh (approved):
lgtm - if you want the checks to pass modify lines 61-62 to add the-f flag in the do.sh to look like this:

```
    && ln -sf /usr/bin/python3 /usr/bin/python \
    && ln -sf /usr/bin/pip3 /usr/bin/pip \
```

flow/flow.yaml
shramroy: @ashutshkumr yes 3 scenarios covered.

ashutshkumr (approved):

Issue #114, opened 2021-05-11
When Setting defaults for a property it is setting that default for "start" and "step" for increment decrement also


shramroy:
When Setting defaults for a property it is setting that default for values & "start" and "step" for increment decrement also.

For Example - 
```yaml
tpid:
          x-field-pattern:
            x-constants:
              x8100: 33024
              x88A8: 34984
              x9100: 37120
              x9200: 37376
              x9300: 37632
            description: >-
              Protocol identifier
            format: integer
            length: 16
            default: 33024
            features: [count, metric_group]
```

this code is generating the below yaml code.

```yaml
Pattern.Flow.Vlan.Tpid.Counter:
      description: |-
        integer counter pattern
      type: object
      required:
      - start
      - step
      properties:
        start:
          type: integer
          default: 33024
          minimum: 0
          maximum: 65535
        step:
          type: integer
          default: 33024
          minimum: 0
          maximum: 65535
        count:
          type: integer
          default: 1
      x-constants:
        x8100: 33024
        x88A8: 34984
        x9100: 37120
        x9200: 37376
        x9300: 37632
    Pattern.Flow.Vlan.Tpid:
      description: |-
        Protocol identifier
      type: object
      required:
      - choice
      properties:
        choice:
          type: string
          enum:
          - value
          - values
          - increment
          - decrement
          default: value
        value:
          type: integer
          default: 33024
          minimum: 0
          maximum: 65535
        values:
          type: array
          items:
            type: integer
            default: 33024
            minimum: 0
            maximum: 65535
        metric_group:
          description: |-
            A unique name is used to indicate to the system that the field may extend the metric row key and create an aggregate metric row for every unique value. To have metric group columns appear in the flow metric rows the flow metric request allows for the metric_group value to be specified as part of the request.
          type: string
        increment:
          $ref: '#/components/schemas/Pattern.Flow.Vlan.Tpid.Counter'
        decrement:
          $ref: '#/components/schemas/Pattern.Flow.Vlan.Tpid.Counter'
      x-constants:
        x8100: 33024
        x88A8: 34984
        x9100: 37120
        x9200: 37376
        x9300: 37632
```

PR #115, opened 2021-05-17
set correct defaults for step and values

ajbalogh:
# Issue #114 
Auto generated schema objects from an x-field-pattern do not have the correct defaults for step and values properties.

# Change
An x-field-pattern will produce generated schema object default keyword/values as follows:
```
if there is a default value
  if format is mac, step default is '00:00:00:00:00:01', values default is ['00:00:00:00:00:01']
  if format is ipv4, step default is '0.0.0.1', values default is ['0.0.0.1']
  if format is ipv6, step default is '::1', values default is ['::1']
  else step default is 1, values defaults is [1]
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+artifacts/openapi.yaml linguist-generated=true
+artifacts/otg.proto linguist-generated=true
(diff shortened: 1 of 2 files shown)


ashutshkumr (approved):
One minor nit, `type` is a kwarg or arg ?
```python
_apply_common_x_field_pattern_properties(self, schema, xpattern, format, type)
```

Rest LGTM.

PR #124, opened 2021-06-10
Support for device ISIS routing protocol

ajbalogh:
[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/ajbalogh/issue119/artifacts/openapi.yaml)

```python
# sample code for configuring a device with ISIS routing protocol enabled
cfg = api.config()
p1 = cfg.port.append(name='p1')
d1 = cfg.devices.append(name='d1')

is = d1.ethernet.isis
is.name = 'i1'
is.interface.system_id = '00000000'
is.interface.metric = 10
```


device/ipv4.yaml
```diff
@@ -36,3 +34,37 @@ components:
...
+    # Can not derive LoopbackIpv4 from Ipv4Base as default prefix length must be 32.
+    Interface.LoopbackIpv4:
+      x-include: 
+      - '../common/common.yaml#/components/schemas/Named.Object'
+      description: >-
+        An IPv4 loopback interface
+      type: object
+      required: [interface_name]
+      properties:
+        interface_name:
```
ajbalogh, line 47: Should this be ethernet_name instead?

device/ipv6.yaml
```diff
@@ -37,3 +35,33 @@ components:
...
+    # Can not derive LoopbackIpv6 from Ipv6Base as default prefix length must be 128.
+    Interface.LoopbackIpv6:
+      x-include: 
+      - '../common/common.yaml#/components/schemas/Named.Object'
+      description: >-
+        An IPv6 loopback interface
+      type: object
+      required: [interface_name]
+      properties:
+        interface_name:
```
ajbalogh, line 48: should this be ethernet_name instead or will the x-constraint be extended in the future?

device/simulatedtopology/simulatedtopology.yaml
```diff
@@ -0,0 +1,127 @@
...
+          maximum: 128
+          default: 64
+        isis_sim_link:
+          description: >-
+              This contains the properties of a simulated Isis Link.
+          $ref: './isissimulatedtopology/isissimulatedtopologylink.yaml#/components/schemas/Router.IsisSimulatedTopology.Link'  
+        ospf_sim_link:
+          description: >-
+            This contains the properties of a simulated Ospf Link.
+          $ref: './ospfsimulatedtopology/ospfsimulatedtopologylink.yaml#/components/schemas/Router.OspfSimulatedTopology.Link'
```
Rangababu-R, line 127: how are these simulated topology is mapped to the router instance?
is it okay to map these under isis and ospf routerange nodes? just like network topology under network group?

ANISH-GOTTAPU:
1) Few places the required field is removed and updated with default values like `eth_mac`, `ipv4_addr`, `gateway`
2) Why two arrays of `bgp_peers` , this will be overhead from the user POV, if one device(router) instance will be having one ipv4 interface, we can't add multiple (bgp_ipv4_peers). Instead of keeping it as a list, can it be a single object?  

```
  r1 = cfg.routers.router(name='r1')[-1]
  bgp = r1.bgp_ipv4_peers.bgpv4peerlist()[-1]
  bgp.ipv4_name = ip1.name
  bgp.bgp_peers.bgpv4(name='bgp1')[-1]
```
routers:
- bgp_ipv4_peers:
  - bgp_peers:
    - active: true
      as_number: 1
      as_number_set_mode: do_not_include_as
      as_number_width: four
      as_type: ibgp
      dut_address: 0.0.0.0
      name: bgp1
      router_id: 0.0.0.0
    ipv4_name: ip1
  name: r1

Rangababu-R (commented):
dividing of interfaces and routers, might introduce client side validation
ex:
router.isis()[-1].interfaces = [eth1.name, eth2.name]
router.isis()[-1].interfaces = [eth2.name, eth3.name]
this is a one erroneous senario which we might end up adding such validation.

device/ipv4.yaml
apratimmukherjee: Yes , this is meant to be the ethernet interface behind which this loopback is connected.

device/ipv6.yaml
apratimmukherjee: Yes, similar to IPv4 Loopback.

apratimmukherjee:
> 1. Few places the required field is removed and updated with default values like `eth_mac`, `ipv4_addr`, `gateway`
> 2. Why two arrays of `bgp_peers` , this will be overhead from the user POV, if one device(router) instance will be having one ipv4 interface, we can't add multiple (bgp_ipv4_peers). Instead of keeping it as a list, can it be a single object?
> 
> ```
>   r1 = cfg.routers.router(name='r1')[-1]
>   bgp = r1.bgp_ipv4_peers.bgpv4peerlist()[-1]
>   bgp.ipv4_name = ip1.name
>   bgp.bgp_peers.bgpv4(name='bgp1')[-1]
> ```
> 
> routers:
> 
> * bgp_ipv4_peers:
>   
>   * bgp_peers:
>     
>     * active: true
>       as_number: 1
>       as_number_set_mode: do_not_include_as
>       as_number_width: four
>       as_type: ibgp
>       dut_address: 0.0.0.0
>       name: bgp1
>       router_id: 0.0.0.0
>       ipv4_name: ip1
>       name: r1

1)  It was not our intention to change fields for Ethernet or IPv4. They should be same as before. I could not find any diffs for the three fields mentioned i.e. `eth_mac`, `ipv4_addr`, `gateway`
2) The intention is to support the scenario where one device has multiple interfaces ( very possible ) and each interface has BGP sessions configured to peer with *multiple* DUTs.   Hence these need to be arrays of arrays to handle both planes .

device/simulatedtopology/simulatedtopology.yaml
apratimmukherjee: Each Router has an array of Simulated Topologies.  Each Router has one ISIS router or one OSPF Router or both.
So the nodes in the ST(s) need to be advertised by either ISIS or OSPF or both. Note that OSPFv3 and LDP can also have properties to advertise in the ST per node/ or per node/link in the future. So this is modelled as a shared object that can be advertised by multiple protocols. It might be very cumbersome to duplicate the same ST twice for OSPF , ISIS and maybe other protocols as well.

apratimmukherjee:
> dividing of interfaces and routers, might introduce client side validation
> ex:
> router.isis()[-1].interfaces = [eth1.name, eth2.name]
> router.isis()[-1].interfaces = [eth2.name, eth3.name]
> this is a one erroneous senario which we might end up adding such validation.

Yes, that is true. 
The question is what we want to support in future. 
1)  Every router can have single interface only ( then how do we emulate even simple use-cases that need a connected interface and a loopback --> advanced use-cases might be one router having 2 interfaces ) 
2)  We decide that this model can support multiple sub-interfaces on same port only . Then array of interface config can be kept  within router but cannot ever have a router with 2 interfaces on 2 different test ports (e.g. backup/active links) i.e. interface  or device list can be removed. 
3) Have set of interfaces per port.
Allow users to have most generic future options while creating routers/devices from these set of interfaces:
a) Multiple routers , each with one connected and one loopback ( with connected = subinterface on some test port.) 
b) Multiple routers , each with multiple connected ,  each connected = subinterface in same test port. ( Possible to support today , supported in some Classic use-cases only) 
c) One router , with 2 interfaces , each interface on a different physical test port. ( Cant model this today in our existing test tools)
     -> if we decide we never want to do this : Then we can have Port --> Router --> [ interface list , isis router ]  i.e. no separate Device or Interface object is needed but in future we should never look for one protocol instance spanning multiple physical test ports ]

winstonliu1111:
converged with #154?

can close?

ajbalogh:
yes it can be closed

winstonliu1111:
closing as we converged on #154

PR #125, opened 2021-06-18
updating models for required fields

ANISH-GOTTAPU:
updating models for required fields and removing defaults

### DIFF
--- a/device/bgp/bgpv4.yaml
+++ b/device/bgp/bgpv4.yaml
@@ -7,7 +7,7 @@ components:
       description: >-
         Container for emulated BGPv4 peers and routes.
       type: object
-      required: [dut_address]
+      required: [local_address, dut_address]
       properties:
         local_address:
           description: >-
--- a/device/bgp/bgpv4.yaml
+++ b/device/bgp/bgpv4.yaml
@@ -7,16 +7,15 @@ components:
       description: >-
         Container for emulated BGPv4 peers and routes.
       type: object
+      required: [dut_address]
       properties:
         local_address:
           description: >-
             Local IPv4 address of the emulated router
           type: string
           format: ipv4
-          default: 0.0.0.0
         dut_address:
           description: >-
             IPv4 address of the BGP peer for the session
           type: string
           format: ipv4
-          default: 0.0.0.0
--- a/device/bgp/bgpv6.yaml
+++ b/device/bgp/bgpv6.yaml
@@ -7,7 +7,7 @@ components:
       description: >-
         Container for BGPv6 peer settings and routes.
       type: object
-      required: [dut_address]
+      required: [local_address, dut_address]
       properties:
         local_address:
           description: >-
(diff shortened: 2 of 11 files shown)


ajbalogh (commented):
Is there a specific issue that is the reason for this change?
Just looking for justification for why defaults are being removed and replaced with required.

ANISH-GOTTAPU:
@ajbalogh : with this PR https://github.com/open-traffic-generator/snappi/pull/72 , if user is not passing the value for properties, defaults are set.

- container_name: port 1
  ethernet:
    ipv4:
      address: 0.0.0.0
      gateway: 0.0.0.0
      name: ipv4
      prefix: 24
    mac: 00:00:fa:ce:fa:ce
    mtu: 1200
    name: eth
  name: device
ports:
- location: 10.39.65.230;4;1
  name: port 1

Earlier, the values were null and ixNetwork used to put ixNetwork defaults, but with new Snappi Version the values are set to `0.0.0.0` which doesn't make much sense. So, we had internal discussion and thought to make them as required and catch them as errors on the client instead of adding not desirable values on the server

device/bgp/bgpv4.yaml
```diff
@@ -7,16 +7,15 @@ components:
       description: >-
         Container for emulated BGPv4 peers and routes.
       type: object
+      required: [dut_address]
       properties:
         local_address:
```
ajbalogh, line 12: removal of default should make this a required property

device/bgp/bgpv6.yaml
```diff
@@ -7,18 +7,17 @@ components:
       description: >-
         Container for BGPv6 peer settings and routes.
       type: object
+      required: [dut_address]
```
ajbalogh, line 10: removal of default should make this a required property

ajbalogh (changes requested):
There are properties that have had the defaults removed, shouldn't these properties should now be required.

device/bgp/bgpv4.yaml
ANISH-GOTTAPU: Most of tgens would take local address from ipv4 stack and populate the local address. removal of default would make it to None and the tgen default values are applied.

device/bgp/bgpv6.yaml
ANISH-GOTTAPU: Most of tgens would take local address from ipv4 stack and populate the local address. removal of default would make it to None and the tgen default values are applied.

ajbalogh: @ANISH-GOTTAPU the comment that concrete implementations would take the local address from the ipv4 stack and apply it to the local address is implicit, the model still requires that property to either have a default value or be required.

device/bgp/bgpv4.yaml
ashutshkumr: `local_address` is now a required field

device/bgp/bgpv6.yaml
ashutshkumr: `local_address` is now a required field

ashutshkumr (approved):
Add some fields to required as discussed.

PR #127, opened 2021-06-22
length add for custom capture filter

shramroy:


capture/capture.yaml
```diff
@@ -83,16 +83,23 @@ components:
             The byte offset to filter on
           type: integer
           default: 0
+        byte_length:
```
ashutshkumr, line 86: change this to bit_length

ashutshkumr (changes requested):

capture/capture.yaml
shramroy: updated accordingly

capture/capture.yaml
```diff
@@ -83,16 +83,23 @@ components:
...
+          type: integer
+          default: 8
     
     Capture.Field:
       type: object
       properties:
         value:
           type: string
           format: hex
+          default: 0
```
ashutshkumr, line 98: should the defaults be "0" or 0 ?

capture/capture.yaml
```diff
@@ -83,16 +83,23 @@ components:
             The byte offset to filter on
           type: integer
           default: 0
+        bit_length:
+          description: >-
+            The bit length to filter on
```
ashutshkumr, line 88: The bit length of field to filter on

ashutshkumr: change the description for byte offset as well. "The byte offset of field to filter on"

capture/capture.yaml
shramroy: In the other fields like mac and ip default was mentioned without quotation. That's why I didn't use quotation.

capture/capture.yaml
shramroy: updated

capture/capture.yaml
shramroy: updated

ashutshkumr (approved):

PR #128, opened 2021-06-25
removing required for choice for x-field-pattern

ANISH-GOTTAPU:


bundler.py
```diff
@@ -232,7 +232,6 @@ def _generate_value_schema(self, xpattern, schema_name, description, type_name,
         schema = {
             'description': description,
             'type': 'object',
-            'required': ['choice'],
```
ashutshkumr, line 235: remove `required` only when `default` is provided

ashutshkumr (changes requested):

ashutshkumr (approved):

Issue #140, opened 2021-07-19
update BGPv4 & BGPv6 metrics response 


ANISH-GOTTAPU:
BGPv4 & BGPv6 metrics response has `sessions_up, sessions_down, sessions_not_started` which doesn't make sense as metrics are per device and one device contains only single session

PR #145, opened 2021-07-22
updating bgp metrics response

ANISH-GOTTAPU:
To address the issue #140

### DIFF
--- a/result/bgpv4.yaml
+++ b/result/bgpv4.yaml
@@ -30,10 +30,7 @@ components:
           items:
             type: string
             enum:
-              - sessions_total
-              - sessions_up
-              - sessions_down
-              - sessions_not_started
+              - session_state
               - routes_advertised
               - routes_withdrawn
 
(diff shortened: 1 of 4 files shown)


result/bgpv4.yaml
```diff
@@ -46,22 +43,10 @@ components:
...
-        sessions_down:
-          description: >-
-            Sessions are not active state
-          type: integer
-        sessions_not_started:
-          description: >-
-            Sessions not able to start due to some internal issue
-          type: integer
+            Session state as up or down
+          type: string
```
ashutshkumr, line 49: this should have enum values "up" and "down"

ashutshkumr (changes requested):

result/bgpv4.yaml
ANISH-GOTTAPU: updated

ashutshkumr (approved):

Issue #143, opened 2021-07-22
Remove defaults for layer1 media


ashutshkumr:
Default is currently set to `copper`, which is does not make sense when working with ixia-c containers or other implementation where we don't have physical nic.

It should default to `null`.

PR #146, opened 2021-07-22
Removing default for layer1 media

indraniBh:


ashutshkumr (approved):

ashutshkumr:
Fixes https://github.com/open-traffic-generator/models/issues/143

Issue #151, opened 2021-08-02
BGP-Neighbors stress and route convergence 


alakendu:
clone this issue from Sonic issue https://github.com/Azure/sonic-mgmt/issues/3611

Protocol level flap(enable/disable) is not available

Changes needed in model/Implementation:
-- add one more state under Control operations (set_protocol_state), need to update the model as well ixNetwork Implementation

PR #152, opened 2021-08-02
Protocol state

alakendu:
This should address https://github.com/open-traffic-generator/models/issues/151

[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/protocol_state/artifacts/openapi.yaml#operation/set_protocol_state)

Code should looks like

```python
import snappi
api = snappi.api(location='http://127.0.0.1:80')
config = api.config()
d = config.devices.device(name='device', container_name=p1.name)[-1]
e = d.ethernet
e.name = 'ether'
e.mac = '00:01:00:00:00:01'
i4 = e.ipv4
i4.name = 'IPv4'
i4.address = '1.1.1.1'
i4.gateway = '1.1.1.2'
b4 = i4.bgpv4
b4.name = 'Bgpv4'
api.set_config(config)

# Current model we start all protocols at the time of start transmit
transmit_state = api.transmit_state()
transmit_state.state = transmit_state.START
api.set_transmit_state(transmit_state)
time.sleep(2)

proto_state = api.protocol_state()
proto_state.names = ['Bgpv4']
proto_state.state = proto_state.STOP
api.set_protocol_state(proto_state)
```


ajbalogh (approved):
lgtm

control/control.yaml
```diff
@@ -95,3 +95,26 @@ components:
...
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Device.Bgpv4/properties/name'
+          - '/components/schemas/Device.Bgpv6/properties/name'
+        state:
+          description: >-
+            Protocol peer specific states
+          type: string
+          enum: [up, down]
```
ankur-sheth, line 120: Is this better described as 'Start' and 'Stop'?  'Up' and 'Down' are the results of the protocol state-machine and also depend on the DuT.

api/api.yaml
```diff
@@ -132,6 +132,26 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/routes:
```
ankur-sheth, line 135: Should this be '/control/routes' or '/control/device' or '/control/routers'?

ankur-sheth (changes requested):

control/control.yaml
apratimmukherjee: I agree that Start and Stop probably make more sense.

apratimmukherjee:
I think we might also need a per port/global  per protocol start/stop.
e.g.  Say we have 1 port with 1000 devices each having one bgp peer and one isis router/session.
And as a first step , I want to start only all the BGP peers .
If we were to use this API, dont we have to fill in the names of all the BGP peers. ( Since looks like ISIS routers will also be using same API and sending empty string will cause all ISIS routers to be started as well ) .

ankur-sheth:
Don't know if we need to solve 1000 routers use-case right now.  Also, how likely is it that one would want to start all BGP but not all ISIS routers?  
If the case is that then can you propose an API that we solves it?

api/api.yaml
alakendu: change it to /control/routers

control/control.yaml
alakendu: change it to Start and Stop

apratimmukherjee:
> Don't know if we need to solve 1000 routers use-case right now. Also, how likely is it that one would want to start all BGP but not all ISIS routers?
> If the case is that then can you propose an API that we solves it?

To start all interfaces and bgp sessions in ports p1 and p2 
proto_state = api.globalProtocolState()
proto_state.names = ['BGP','INTERFACES']
proto_stat.port_names= ['p1','p2']
proto_state.state = proto_state.START  ( or .STOP  to stop selected protocols on selected ports ) 
api.set_global_protocol_state(proto_state)

To start all protocols on all ports 
proto_state = api.globalProtocolState()
proto_state.names = []
proto_stat.port_names= []
proto_state.state = proto_state.START  ( or .STOP  to stop all protocols on all ports ) 
api.set_global_protocol_state(proto_state)

To start only BGP on all ports 
proto_state = api.globalProtocolState()
proto_state.names = ['BGP']
proto_stat.port_names= []
proto_state.state = proto_state.START  ( or .STOP  to stop all protocols on all ports ) 
api.set_global_protocol_state(proto_state)

ankur-sheth:
@apratimmukherjee Approach seems ok.  Some specific questions:

1. What if I only want to start 'bgp1' and 'bgp2', while I also have 'bgp3' configured.  Again, not sure if we need to support such case and add it to the model right away or not, but would be good to have some kind of clarity on our preferred approach.
2. Is `proto_state.names` more appropriately named as `proto_state.protocols` and `proto_state.names` used for the specific instances (ie "BGP1", "my-isis3", etc.)?

winstonliu1111:
where are we with this PR?  hasn't got response since early August.  @apratimmukherjee any comment to Ankur's question?  we should incorporate this along with the new protocols model changes to not introduce another round of breakage in behavior.

ankur-sheth:
Agree.  This needs to go in along with the model restructuring.  Once we are at 0.6.1, we cannot do backwards incompatible changes anymore.

apratimmukherjee:
> where are we with this PR? hasn't got response since early August. @apratimmukherjee any comment to Ankur's question? we should incorporate this along with the new protocols model changes to not introduce another round of breakage in behavior.
Sorry , missed this. I think it is difficult to mix global stop/start with control for individual protocols and start/stop of specific peers or isis/ospf routers. e.g in protocol list you have BGP and ISIS , in proto_state.names we have bgp peer1, bgp peer 3  and in port list we have p1,p2  then will it mean that all isis and 2 bgp routers ?  What if bgp peer1 and bgp peer 3 is on p3 ?  We end up having three sets of data , protocols , ports and specific routers/peers. For me specific peers/routers are probably handled better in a separate API which looks like above. Global protocol  state start/stop can have only protocols/ports and peer/router specific  protocol_state API has only specific peers/routers  and state.  Implementation can choose to support one or both APIs . We can start with any one of them. Start All looks easier to implement. Start specific peer/router is much more complex since we need to start just that and  its children .  Also start /stop /start for specific or peers/routers means that the port will probably first have to store a copy of the full setconfig data so not sure if it can be implemented very quickly. Currently we just transfer the SetConfig data to the daemon and it cleans that on next stop , so next start means we have to hold on to a copy of the SetConfig data. We can implement that but will need some time to create the heirarchial config database.

winstonliu1111:
@ANISH-GOTTAPU @Rangababu-R @ashutshkumr plans to drive this forward?

api/api.yaml
```diff
@@ -132,6 +132,26 @@ paths:
...
+        Updates the state of the device protocol peer.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../control/control.yaml#/components/schemas/Protocol.State'
+      responses:
+        '200':
+          $ref: '../result/request.yaml#/components/responses/Success'
```
ajbalogh, line 149: The response is not flexible enough to indicate what "names" failed to start/stop.  The response should be more detailed to allow users to determine if they want the test to continue or fail in the case where the protocol state isn't met.

ajbalogh (changes requested):

api/api.yaml
ashutshkumr: this makes more sense if regex is used in request names.

ashutshkumr:
Closing because this is and opening https://github.com/open-traffic-generator/models/pull/169

ByReaL:
tried a test recently for a port monitor scenario, i have 3 ports in test on 2 of them protocols are up and on third one protocols are down on purpose, (bgp not the link) i need to start the traffic with bgp up on 2 ports and down on the third port but i cannot do that because is waiting for all protocols to be up.

how can we star traffic with protocols down ?

Issue #142, opened 2021-07-21
Should update_config be used only for OTF (on the fly) changes


rudranil-das:
These questions came up in a ixia-c internal discussion and following are thoughts that requires validation from
model perspective.

a. How user is expected to use set_config and update_config when the constructs of two are same
   - purpose of update_config is to have a quicker turn-around time for pushing configurations to traffic-engine,
     compared to set_config i.e. the traffic-engine can accommodate the changed items without stopping
     current execution of flows/devices which points to OTF changes from traffic-engine perspective.
   - the configurations which can be applied OTF to traffic-engine should be allowed by controller implementation
     in update_config; rest should be prohibited through validation.

b. Should there be any change in model to identify/restrict non OTF applicable changes
   - No, because OTF applicable changes would vary from one traffic-engine to other and should be taken
     care of in controller implementation.

c. Should there be both set_config(POST) and update_config (PATCH) or just set_config(POST) with an enum set/update
   - this does not have as such implication on controller implementation
   - current thought is yes, we need to keep both.

rudranil-das:
@ankur-sheth @ajbalogh @ashutshkumr @arkajyoti-cloud

ashutshkumr:
@winstonliu1111

ajbalogh:
Some thoughts for discussion: update_config has not been implemented as there has been no specific ask. This is the first opportunity to fully specify the model specifics for this method.  The intent is to update a config with only changes. The implementation should determine what changes should be applied as the patch content can be items that are being added, updated, or deleted.

winstonliu1111:
i don't know the right solution to this.  in general, update_config kinda distracts the message of "declarative state/configs", it sort of bring back imperative semantics back into config.  

maybe we need to delineate based on:
1) if some kind of containers in the config need to be added/deleted (that should really be through set config, and controller should do its best to quickly detect diffs and apply the diff)
2) "actions/operations that you want to enact on a particular object or set of objects, or its properties. (think for example, changing rates of a stream dynamically is reasonable OTF change, and fit this category of (2).

maybe look at how openconfig delineates between gnmi and gnoi....maybe there is a parallel here for on the fly change on already set config?

ajbalogh:
gnoi proto for bgp https://github.com/openconfig/gnoi/blob/master/bgp/bgp.proto

ashutshkumr:
Per recent discussions, here are the three approaches:

PATCH /config
```python
	{
		# reuse config and modify wherever needed
[...]

PR #155, opened 2021-08-12
Otf rate

rudranil-das:
This PR is intended to fix issue [142](https://github.com/open-traffic-generator/models/issues/142) based on the discussion and last proposed model by Andy.

[Please view the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/otf_rate/artifacts/openapi.yaml#operation/update_flow_state)

A sample snippet should look like

```
import snappi
api = snappi.api(location='https://127.0.0.1:443')

# Create Config
config = api.config()
tx, rx = (
    config.ports
    .port(name='tx', location="localhost:5555")
    .port(name='rx', location="localhost:5556")
)

ly = config.layer1.layer1(name='l1')[0]
ly.speed = ly.SPEED_1_GBPS
ly.port_names = [tx.name, rx.name]

f1, f2, f3 = config.flows.flow(name='flow1').flow(
    name='flow2').flow(name='flow3')

for flow in [f1, f2, f3]:
    flow.packet.ethernet()
    flow.size.fixed = 128
    flow.rate.mbps = 10
    flow.tx_rx.port.tx_name = tx.name
    flow.tx_rx.port.rx_name = rx.name

# Set Config
api.set_config(config)

# Start transmit
ts = api.transmit_state()
ts.state = ts.START
api.set_transmit_state(ts)

# Update flows
fupd = api.flow_updates()

f1.size.fixed = 64
fupd.append(f1.name, f1.size)

f2.rate.pps = 100
fupd.append(f2.name, f2.rate)

f3.size.fixed = 100
f3.rate.percentage = 50
fupd.append(f3.name, f3.size, f3.rate)

api.update_flow_state(fupd)
```

### DIFF
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -97,9 +97,9 @@ paths:
   /control/flows:
     post:
       tags: ['Control']
-      operationId: update_flow_state
+      operationId: update_flows
       description: >-
-        Updates the state of configuration resources on the traffic generator.
+        Updates flow properties without disruption of transmit state.
       requestBody:
         required: true
         content:
@@ -109,7 +109,7 @@ paths:
       responses:
         '200':
           description: >-
-            Response with current Config from the traffic generator
+            Response with updated Config from the traffic generator
           content:
             application/json:
               schema:
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -94,6 +94,31 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/flows:
+    post:
+      tags: ['Control']
+      operationId: update_flow_state
+      description: >-
+        Updates the state of configuration resources on the traffic generator.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../control/control.yaml#/components/schemas/Flow.Update'
+      responses:
+        '200':
+          description: >-
+            Response with current Config from the traffic generator
+          content:
+            application/json:
+              schema:
+                $ref: '../config/config.yaml#/components/schemas/Config'
+        '400':
+          $ref: '../result/request.yaml#/components/responses/BadRequest'
+        '500':
+          $ref: '../result/request.yaml#/components/responses/InternalServerError'
+
   /control/routes:
     post:
       tags: ['Control']
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -105,7 +105,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '../control/control.yaml#/components/schemas/Flow.Update'
+              $ref: '../control/control.yaml#/components/schemas/Flows.Update'
       responses:
         '200':
           description: >-
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -18,24 +18,6 @@ paths:
           $ref: '../result/request.yaml#/components/responses/BadRequest'
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
-    patch:
-      tags: ['Configuration']
-      operationId: update_config
-      description: >-
-        Updates configuration resources on the traffic generator.
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '../config/config.yaml#/components/schemas/Config'
-      responses:
-        '200':
-          $ref: '../result/request.yaml#/components/responses/Success'
-        '400':
-          $ref: '../result/request.yaml#/components/responses/BadRequest'
-        '500':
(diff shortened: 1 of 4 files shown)


ajbalogh (changes requested):
In the main comment there is no link to the proposed changes using redocly (this allows viewing the model in its entirety).
In the main comment there is no sample snappi code and/or json snippet that shows the usage of the proposed model change.

control/control.yaml
```diff
@@ -73,6 +73,29 @@ components:
...
           enum: [start, stop]
 
+    Flow.Update:
+      description: A container for properties that define flow property changes
+      type: object
+      properties:
+        update:
+          description: Indicates what child nodes of a flow are being updated on the fly. 
+          type: array
+          items:
```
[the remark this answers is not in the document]
rudranil-das: @ajbalogh , I have updated the PR with your change-request.
Also updated the main comment with redocly-viewer link and sample code-snippet

ashutshkumr:
I like the fact that we return updated `Config` back. But request seems like something which will require more work on the user's side.

Shouldn't we just be passing a list of flows and indicate the name of fields we wish to update ? e.g. something like this

```python
# contains one two flows with fixed frame size 64 and rate 100 pps
config = api.config()
api.set_config(config)

# update frame size to 68 and rate to 1000 pps
for f in config.flows:
    f.size.fixed = 68
    f.rate.pps = 1000

req = api.flow_state_request()
req.flows = config.flows
req.update = [req.SIZE, req.RATE]

config = api.update_flow_state(req)
```

The snippet may not be entirely correct.

rudranil-das:
The format proposed by @ashutshkumr in last [comment](https://github.com/open-traffic-generator/models/pull/155#issuecomment-901988601) is reviewed and considered as a better option by @winstonliu1111 and @ankur-sheth . I would update the changes soon.

api/api.yaml
```diff
@@ -112,6 +94,31 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/flows:
+    post:
+      tags: ['Control']
```
ashutshkumr, line 99: shouldn't it belong to Config tag instead of Control ? since we're essentially modifying flow properties.

api/api.yaml
```diff
@@ -112,6 +94,31 @@ paths:
...
+      operationId: update_flow_state
```
ashutshkumr, line 100: flow state may be a little ambiguous; maybe we should just call it `update_flows`

api/api.yaml
```diff
@@ -112,6 +94,31 @@ paths:
...
+      description: >-
+        Updates the state of configuration resources on the traffic generator.
```
ashutshkumr, line 102: this description is very generic. we should mention that it updates flow properties without disrupting transmit state.

api/api.yaml
```diff
@@ -112,6 +94,31 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../control/control.yaml#/components/schemas/Flows.Update'
+      responses:
+        '200':
+          description: >-
+            Response with current Config from the traffic generator
```
ashutshkumr, line 112: current -> updated

control/control.yaml
```diff
@@ -73,6 +73,23 @@ components:
           type: string
           enum: [start, stop]
 
+    Flows.Update:
+      description: A container of flows with associated property updates intended for on the fly change
```
ashutshkumr, line 77: although we can mention "on the fly" in brackets, I think we should be more explicit by saying flow properties will be updated without affecting the transmit state

control/control.yaml
```diff
@@ -73,6 +73,23 @@ components:
...
+      type: object
+      required: [update]
+      properties:
+        update:
```
ashutshkumr, line 81: update -> property_names

control/control.yaml
```diff
@@ -73,6 +73,23 @@ components:
...
+          description: Flow properties to be updated on the fly 
+          type: string
+          enum: [rate, size]
+        flows:
+          description: >-
+            The list of configured flows on which update will be applied.
```
ashutshkumr, line 87: The list of configured flows for which given properties will be updated

ashutshkumr (changes requested):

api/api.yaml
rudranil-das: By looking at the existing tags (esp. with reference to /control/link, /control/routes), my understanding is tag is more associated with classification of type of api and not based on its operands - is it not ?
From that perspective , I have kept it as Control which seems in line.

ashutshkumr (approved):

PR #157, opened 2021-08-18
Flow.tx_rx tx_names and rx_names

ajbalogh:


flow/endpoint.yaml
```diff
@@ -70,6 +70,15 @@ components:
           enum: [mesh, one_to_one]
           default: mesh
         tx_names:
+          description: |-
+            The system will use information from the supplied tx_names to populate the following
+            flow.packet.header fields if they are present:
+              - ethernet.dst using device ipv4 addresses from either interfaces, route ranges etc.
+              - ethernet.src using ethernet.mac addresses from ethernet interfaces
+              - ipv4.src using device ipv4 addresses from either interfaces, route ranges etc.
```
vinjose, line 78: Do we need to document this now -  Those are convenience features that can be looked at later.

ajbalogh: Those implicit actions are already being done in other implementations (snappi[ixnetwork]) so they should be documented now so there is no confusion at a later date.

ajbalogh: And users of snappi[ixnewtork] are already expecting that behavior.

ashutshkumr:
Although no change was done for txrx port, I think following description needs to be changed:
"A container for a transmit port and 0..n intended receive ports" -> should be 0..1 intended receive port.

ashutshkumr:
For one_to_one mapping, here’s the excerpt from documentation.
 
> The one_to_one mode creates traffic between each value pair in of tx_names and rx_names by index. The length of tx_names and rx_name MUST be the same. Identical values in tx_names or rx_names MUST raise an error.
 
I see one exception here, i.e. the document does not explicitly state what happens if number of addresses in bgpv4 route rr1 (tx) is different from rr2 (rx).

We don't need to use route range terminology, but we need to specify that length of values contained in each corresponding device needs to be same as well.

flow/endpoint.yaml
```diff
@@ -70,6 +70,15 @@ components:
           enum: [mesh, one_to_one]
           default: mesh
         tx_names:
+          description: |-
+            The system will use information from the supplied tx_names to populate the following
+            flow.packet.header fields if they are present:
+              - ethernet.dst using device ipv4 addresses from either interfaces, route ranges etc.
```
ashutshkumr, line 76: shouldn't this description be in rx_names  and modified a little to imply "ethernet.dst using resolved mac addresses obtained using device ipv4 addresses from either interfaces, route ranges etc." ?

PR #158, opened 2021-08-22
Fix update_flows api

rudranil-das:
Proposed changes:
1. Change _property_names_ to _property_name_ as a single property is updated at a time.
2. Make the argument _flows_ mandatory because with only _property_name_ enum value and nil/empty _flows_ the operation becomes NO-OP for any implementation, because the updated property value(s) are supposed to be contained in one or more of the flows in _flows_ array. 

[Please view the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/otf_rate_fix/artifacts/openapi.yaml#operation/update_flows)

### DIFF
--- a/control/control.yaml
+++ b/control/control.yaml
@@ -74,18 +74,17 @@ components:
           enum: [start, stop]
 
     Flows.Update:
-      description: A container of flows with associated properties to be updated without affecting the transmit state
+      description: A container of flows with associated property to be updated without affecting the transmit state
       type: object
-      required: [property_names]
+      required: [property_name, flows]
       properties:
-        property_names:
-          description: Flow properties to be updated without affecting the transmit state 
+        property_name:
+          description: Flow property to be updated without affecting the transmit state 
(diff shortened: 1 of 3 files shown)


control/control.yaml
```diff
@@ -74,18 +74,17 @@ components:
...
     Flows.Update:
-      description: A container of flows with associated properties to be updated without affecting the transmit state
+      description: A container of flows with associated property to be updated without affecting the transmit state
       type: object
-      required: [property_names]
+      required: [property_name, flows]
       properties:
-        property_names:
-          description: Flow properties to be updated without affecting the transmit state 
+        property_name:
```
ashutshkumr, line 81: shouldn't property_names be an array of enums ?

rudranil-das: updated (and retained name as _property_names_)

ashutshkumr (approved):

PR #161, opened 2021-08-27
device_names -> peer_names, additions of more stats

SuryyaKrJana:


result/bgpv4.yaml
```diff
@@ -24,35 +24,89 @@ components:
...
           description: >-
-            Session state as up or down
+            Session state as up or down.
           type: string
           enum: [up, down]
-        routes_advertised:
+        flap_count:
           description: >-
-            Number of advertised routes sent
+            Number of times the session went from Up to Down state.
```
ashutshkumr, line 63: Should be rephrased as: Number of times session_state changed from up to down.

result/bgpv6.yaml
```diff
@@ -25,34 +25,89 @@ components:
...
           description: >-
-            Session state as up or down
+            Session state as up or down.
           type: string
           enum: [up, down]
-        routes_advertised:
+        flap_count:
           description: >-
-            Number of advertised routes sent
+            Number of times the session went from Up to Down state.
```
ashutshkumr, line 64: Should be rephrased as: Number of times session_state changed from up to down.

result/bgpv4.yaml
```diff
@@ -24,35 +24,89 @@ components:
...
       properties:
         name:
           description: >-
-            The name of a configured BGPv4 device.
+            The name of a configured BGPv4 peer.
           type: string
         session_state:
           description: >-
-            Session state as up or down
+            Session state as up or down.
```
ashutshkumr, line 58: Since we're on it, I think it would be useful to provide more details on what `up` or `down` stands for here.

i.e. my understanding is `up` refers to `Established` state, and `down` refers to others. we of course, might need to rephrase a little more.

result/bgpv4.yaml
SuryyaKrJana: we can but ideally comments should not dependent on variable/function name.

result/bgpv6.yaml
SuryyaKrJana: we can but ideally comments should not dependent on variable/function name.

result/bgpv4.yaml
SuryyaKrJana: yes. we can

result/bgpv4.yaml
```diff
@@ -24,35 +24,89 @@ components:
...
             type: string
             enum:
               - session_state
-              - routes_advertised
-              - routes_withdrawn
+              - flap_count
+              - routes_sent
+              - routes_received
+              - route_withdraws_sent
+              - route_withdraws_received
```
ashutshkumr, line 37: is there a difference between `routes_sent` and `route_advertisements_sent` ?

if latter is more appropriate, then we should be consistent with naming; route_<plural noun>_<sent/received>
- route_advertisements_sent
- route_advertisements_received
- route_withdrawals_sent
- route_withdrawals_received

result/bgpv4.yaml
```diff
@@ -24,35 +24,89 @@ components:
...
+              - opens_received
+              - keepalives_sent
+              - keepalives_received
+              - notifications_sent
+              - notifications_received
 
     Bgpv4.Metric:
       description: >-
-        BGPv4 Router statistics and learned routing information
+        BGPv4 per peer statistics information.
```
ashutshkumr, line 49: Rephrase to: Per peer metrics for BGPv4.
This needs to be fixed for BGPv6 too.

Please remove all occurrences of `statistics` and `information`

ashutshkumr (changes requested):

result/bgpv4.yaml
ashutshkumr: I'm sorry, I didn't understand this part.

I'm actually asking here to refer to actual variable name (`session_state`) and its actual values (`up` or `down`).

result/bgpv4.yaml
```diff
@@ -24,35 +24,89 @@ components:
...
-            The name of the BGPv4 device cannot be excluded.
+            any result_groups. The name of the BGPv4 peer cannot be excluded.
           type: array
           items:
             type: string
             enum:
               - session_state
-              - routes_advertised
-              - routes_withdrawn
+              - flap_count
```
ashutshkumr, line 33: in future, will we ever have something to count route flaps (instead of session flaps) ?

If so, I think calling this `session_flaps` makes more sense.

SuryyaKrJana: No plan to add route flap count for now. We will only count session flaps and "session_flap_count" is more self explantory name.

result/bgpv4.yaml
SuryyaKrJana: Second does not look right due to confusion between routes_withdrawn and routes_removed/deleted. First one is closer to what we want but routes_sent feels more appropriate since "advertisements" and "sent" have similar meaning. Hence the short name is better.

result/bgpv4.yaml
SuryyaKrJana: We feel statistics is more generic term to state about current status of a peer.

result/bgpv4.yaml
SuryyaKrJana: Every description should be indeppendent of others as far possible.

ashutshkumr (approved):

PR #162, opened 2021-09-08
Change capture overwrite to true

shramroy:


ankur-sheth:
Hi guys, can you please make sure that you add a proper description in the PR and the reason for doing this?  Seems to me that this is going to be backwards incompatible.  I'm not averse to making a backwards incompatible change, but the rationale needs to be clearly spelled out.

shramroy:
> Hi guys, can you please make sure that you add a proper description in the PR and the reason for doing this? Seems to me that this is going to be backwards incompatible. I'm not averse to making a backwards incompatible change, but the rationale needs to be clearly spelled out.

The reason for making this change is overwriting capture is not supported in our controller right now. The limit of “capture buffer” is 1 mb. So the capture is being overwritten by default in the controller side. Since model and controller was not in sync we made the default to “true” i.e capture will be overwritten by default. 

It won’t change anything as earlier also we used to overwrite the capture. 

It will now throw a warning if someone tries to set overwrite “false”. 

I will update the description of the PR.

ankur-sheth:
Ok.  Did we check with the snappi[ixnetwork] team and/or Andy/Ashutosh to confirm that they are ok with this change?

shramroy:
> Ok. Did we check with the snappi[ixnetwork] team and/or Andy/Ashutosh to confirm that they are ok with this change?

@ashutshkumr agreed with the change. Mentioning @ajbalogh to have a look.

PR #163, opened 2021-09-08
Support for Ping

ANISH-GOTTAPU:
[Please view the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/ping/artifacts/openapi.yaml#operation/update_flow_state)

The below is the example script for ping requests from user's perspective
 
```
def test_ping(api, b2b_raw_config, utils):
    """
    Demonstrates test to send ipv4 and ipv6 pings

    Return the ping responses and validate as per user's expectation
    """
    port1, port2 = b2b_raw_config.ports
    d1, d2 = b2b_raw_config.devices.device(
        name="tx_bgp").device(name="rx_bgp")
    d1.container_name, d2.container_name = port1.name, port2.name
    eth1, eth2 = d1.ethernet, d2.ethernet
    eth1.mac, eth2.mac = "00:00:00:00:00:11", "00:00:00:00:00:22"
    ip1, ip2 = eth1.ipv4, eth2.ipv4
    ipv61, ipv62 = eth1.ipv6, eth2.ipv6
    eth1.name, eth2.name = "eth1", "eth2"
    ip1.name, ip2.name = "ip1", "ip2"
    ipv61.name, ipv62.name = "ipv6-1", "ipv6-2"

    ip1.address = "10.1.1.1"
    ip1.gateway = "10.1.1.2"
    ip1.prefix = 24

    ip2.address = "10.1.1.2"
    ip2.gateway = "10.1.1.1"
    ip2.prefix = 24

    ipv61.address = "3000::1"
    ipv61.gateway = "3000::2"
    ipv61.prefix = 64

    ipv62.address = "3000::2"
    ipv62.gateway = "3000::1"
    ipv62.prefix = 64

    api.set_config(b2b_raw_config)

    utils.start_traffic(api, b2b_raw_config, start_capture=False)

    try:
        req = api.ping_request()
        p1, p2, p3, p4 = req.endpoints.ipv4().ipv4().ipv6().ipv6()
        p1.src_name = ip1.name
        p1.dst_ip = "10.1.1.2"
        p2.src_name = ip1.name
        p2.dst_ip = "10.1.1.3"
        p3.src_name = ipv62.name
        p3.dst_ip = "3000::1"
        p4.src_name = ipv62.name
        p4.dst_ip = "3000::9"

        responses = api.send_ping(req).responses
        for resp in responses:
            if resp.src_name == ip1.name and resp.dst_ip == "10.1.1.2":
                assert resp.result == "success"
            elif resp.src_name == ip1.name and resp.dst_ip == "10.1.1.3":
                assert resp.result == "failure"
            elif resp.src_name == ipv62.name and resp.dst_ip == "3000::1":
                assert resp.result == "success"
            elif resp.src_name == ipv62.name and resp.dst_ip == "3000::9":
                assert resp.result == "failure"
    except Exception as e:
        utils.stop_traffic(api, b2b_raw_config)
        raise Exception

```


ajbalogh (changes requested):
The model is limiting in that only a single ping can be sent at any time, if there were multiple sources, multiple calls would have to be made. The model should be flexible enough to take as many ping sources, with each ping source having multiple destinations. The result should be flexible enough to grow over time which makes it a Ping.Result object. This allows the result to accomodate any ping content a user may want even though it is not yet outlined.

api/api.yaml
```diff
@@ -139,6 +139,33 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              $ref: '../control/control.yaml#/components/schemas/Ping.Request'
+      responses:
+        '200':
+          description: >-
+            Capture ping response from the traffic generator. 
+          content:
+            application/json:
```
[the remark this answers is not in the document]
ANISH-GOTTAPU: updated

ANISH-GOTTAPU:
> The model is limiting in that only a single ping can be sent at any time, if there were multiple sources, multiple calls would have to be made. The model should be flexible enough to take as many ping sources, with each ping source having multiple destinations. The result should be flexible enough to grow over time which makes it a Ping.Result object. This allows the result to accomodate any ping content a user may want even though it is not yet outlined.

@ajbalogh : updated as per the comment, I have added ping.yaml under control folder..please let me know if I have to create one more folder for ping and move the file there

control/control.yaml
```diff
@@ -113,3 +113,15 @@ components:
             Route specific states
           type: string
           enum: [withdraw, advertise]
+
+    Ping.Request:
+      description: >-
+        Ping request details
+      type: object
+      properties:
+        ping:
```
ajbalogh, line 122: This should be plural 'pings'

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+              destination address
+            type: string
+            format: ipv4
+    
+    Ping.Ipv6:
+      type: object
+      properties:
+        src_name:
+          description: >-
+            The name of source to send ping.
```
ajbalogh, line 48: Make this description similar to the one for ipv4 src_name

ajbalogh (changes requested):

control/control.yaml
ANISH-GOTTAPU: changed to pings

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+    All ping schemas
+  version: ^0.0.0
+
+components:
+  schemas:
+    Ping:
+      type: object
+      properties:
+        choice:
+          description: ipv4 or ipv6 ping
```
ankur-sheth, line 15: IPv4 or IPv6 ping.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+      properties:
+        src_name:
+          description: >-
+            A base IPv4 interface
+          type: string
+          x-constraint:
+          - '/components/schemas/Device.Ipv4Base/properties/name'
+        dst_ips:
+          description: >-
+              Array of destination ipv4 addresses
```
ankur-sheth, line 35: Array of IPv4 addresses to ping.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+      properties:
+        src_name:
+          description: >-
+            A base IPv6 interface
+          type: string
+          x-constraint:
+          - '/components/schemas/Device.Ipv6Base/properties/name'
+        dst_ips:
+          description: >-
+              Array of destination ipv6 addresses
```
ankur-sheth, line 54: Array of IPv6 addresses to ping.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Device.Ipv4Base/properties/name'
+        dst_ips:
+          description: >-
+              Array of destination ipv4 addresses
+          type: array
+          items:
+            description: >-
+              destination address
```
ankur-sheth, line 39: Destination address.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Device.Ipv6Base/properties/name'
+        dst_ips:
+          description: >-
+              Array of destination ipv6 addresses
+          type: array
+          items:
+            description: >-
+              destination address
```
ankur-sheth, line 58: Destination address.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+            type: string
+            format: ipv6
+    
+    Ping.Response:
+      description: >-
+        A container for the different ping responses.
```
ankur-sheth, line 64: A container for ping responses.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+          type: array
+          items:
+            $ref: '#/components/schemas/Response'
+
+    Response:
+      type: object
+      properties:
+        src_name:
+          description: >-
+            The name of the source the ping is requested
```
ankur-sheth, line 77: The name of the source IPv4 or IPv6 interface from which ping was sent.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+          type: string
+          x-constraint:
+            - '/components/schemas/Device.Ipv4Base/properties/name'
+            - '/components/schemas/Device.Ipv6Base/properties/name'
+        dst_ip:
+          description: >-
+            destination address
```
ankur-sheth, line 84: Destination address.

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+          type: string
+        result:
+          description: >-
+            destination address
```
ankur-sheth, line 88: Result of the ping request.

api/api.yaml
```diff
@@ -139,6 +139,32 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/ping:
```
ankur-sheth, line 142: I'm presuming here that the API will wait for a certain amount of time (might even be 50ms) before returning.  From the description it is not clear how long the API will wait.  
Also, it is assumed that only one ping packet will be sent?

api/api.yaml
```diff
@@ -139,6 +139,32 @@ paths:
...
+    post:
+      tags: ['Control']
+      operationId: send_ping
+      description: >-
+        Control operation to send ping
```
ankur-sheth, line 147: API to send an IPv4 and/or IPv6 ICMP Echo Request(s).

ankur-sheth (changes requested):

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

control/ping.yaml
ANISH-GOTTAPU: updated

api/api.yaml
ANISH-GOTTAPU: As of now we are just returning if ping is a success or failed, right now it depends on the default time as per the implementation on particular product.

api/api.yaml
ANISH-GOTTAPU: updated

api/api.yaml
ashutshkumr: - for each src/dst interface pair, ping packet count is 1 and API shall wait for ping response to either be successful or timeout
- we can't provide duration of successful response because RTT for ping may vary based on location of destination interface and number of src/dst interface pairs may be more than 1
- we should however provide an info on how long should it wait before timing out (this could be configurable in future)

@ANISH-GOTTAPU could you please update description accordingly ?

control/ping.yaml
```diff
@@ -0,0 +1,93 @@
...
+  description: >-
+    All ping schemas
+  version: ^0.0.0
+
+components:
+  schemas:
+    Ping:
+      type: object
+      properties:
+        choice:
```
ajbalogh, line 14: Is choice needed here?
A user should be allowed to request ipv4 and ipv6 pings to be sent in the same call.
The response is flexible enough to allow for ipv4 and/or ipv6.

ajbalogh (changes requested):

ashutshkumr:
Currently there's a one-to-many mapping in ping request and one-to-one mapping in ping response, for src and dst.

It will be much simpler, if we could do something like this:

```python

req = api.ping_request()
p1, p2, p3 = req.pairs.pair().pair().pair()
p1.v4.src_name = 'ip1'
p1.v4.dst_ip = '1.1.1.1'
p1.v4.src_name = 'ip2'
p1.v4.dst_ip = '1.1.1.2'
p1.v6.src_name = 'ip3'
p1.v6.dst_ip = '::1'

res = api.send_ping(r)
assert all([r.result == r.SUCCESS for r in res]) and len(res) == 3
```

control/ping.yaml
ANISH-GOTTAPU: I have added the choice for v4 & v6 so that we can do client side validation for IPv4 & IPv6 address. If we make them in same call, we might miss on that and need to add the validation at the server. Please let me know if removing choice is more better option, I will go with it.

api/api.yaml
ankur-sheth: We don't want API behavior to be dependent on the implementation of a particular product.
I need to know from the API description the following a) # of ping packets sent, and b) timeout value - ie the time the implementation waits before declaring the request as failed.  We can choose not to expose the timeout value and the description can reflect the current implementation.

ankur-sheth (changes requested):
Description needs to be updated.

ashutshkumr (approved):
Nit: Change `failed` -> `failure` (since success is a noun)

ankur-sheth (approved):

ajbalogh (approved):
lgtm - the ability to specify ping by discrete names can be achieved on the client side by offering find/regex utilities on the json - the model shouldn't chase individual cases.

PR #167, opened 2021-09-17
Model guide update

ankur-sheth:


ankur-sheth:
I think this branch was created off the wrong base.   Will redo.

ankur-sheth:
Deleting the wrong branch.

PR #169, opened 2021-09-21
Exposed MVP API for toggling protocol state

ashutshkumr:
This should partially address https://github.com/open-traffic-generator/models/issues/151

[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/protocol-state/artifacts/openapi.yaml)

Limitations:
- Currently protocol state toggling is not supported for subset of devices
- Currently the response only returns warning and error strings (the existing response will be extended to include `names` for which operation succeeded and failed)

Code should looks like

```python
import snappi
api = snappi.api(location='http://127.0.0.1:80')
config = api.config()
d = config.devices.device(name='device', container_name=p1.name)[-1]
e = d.ethernet
e.name = 'ether'
e.mac = '00:01:00:00:00:01'
i4 = e.ipv4
i4.name = 'IPv4'
i4.address = '1.1.1.1'
i4.gateway = '1.1.1.2'
b4 = i4.bgpv4
b4.name = 'Bgpv4'
api.set_config(config)

# start protocol sessions
proto_state = api.protocol_state()
proto_state.state = proto_state.START
api.set_protocol_state(proto_state)

transmit_state = api.transmit_state()
transmit_state.state = transmit_state.START
api.set_transmit_state(transmit_state)
```

Closed https://github.com/open-traffic-generator/models/pull/152 as it was way behind HEAD.

### DIFF
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -172,7 +172,7 @@ paths:
       tags: ['Control']
       operationId: set_protocol_state
       description: >-
-        Sets all configured protocols to intended state
+        Sets all configured protocols to `start` or `stop` state.
       requestBody:
         required: true
         content:
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -167,6 +167,25 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/protocols:
+    post:
+      tags: ['Control']
+      operationId: set_protocol_state
+      description: >-
+        Sets all configured protocols to intended state
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../control/control.yaml#/components/schemas/Protocol.State'
+      responses:
+        '200':
+          $ref: '../result/request.yaml#/components/responses/Success'
+        '400':
+          $ref: '../result/request.yaml#/components/responses/BadRequest'
+        '500':
+          $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
   /results/metrics:
     description: >-
--- a/artifacts/openapi.yaml
+++ b/artifacts/openapi.yaml
@@ -189,7 +189,7 @@ paths:
       - Control
       operationId: set_protocol_state
       description: |-
-        Sets all configured protocols to intended state
+        Sets all configured protocols to `start` or `stop` state.
       requestBody:
         required: true
         content:
@@ -3270,7 +3270,7 @@ components:
             $ref: '#/components/schemas/Ping'
     Protocol.State:
       description: |-
-        Sets all configured protocols to intended state
+        Sets all configured protocols to `start` or `stop` state.
       type: object
       required:
       - state
--- a/artifacts/otg.proto
+++ b/artifacts/otg.proto
@@ -3156,7 +3156,7 @@ message PingRequest {
 }
 
 message ProtocolState {
-  option (msg_meta).description = "Sets all configured protocols to intended state";
+  option (msg_meta).description = "Sets all configured protocols to `start` or `stop` state.";
 
   message State {
     enum Enum {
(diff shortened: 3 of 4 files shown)


api/api.yaml
```diff
@@ -167,6 +167,25 @@ paths:
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
 
+  /control/protocols:
+    post:
+      tags: ['Control']
+      operationId: set_protocol_state
+      description: >-
+        Sets all configured protocols to intended state
```
apratimmukherjee, line 175: started or stopped ?

artifacts/otg.proto
```diff
@@ -12078,6 +12103,9 @@ service Openapi {
   rpc SendPing(SendPingRequest) returns (SendPingResponse) {
     option (rpc_meta).description = "API to send an IPv4 and/or IPv6 ICMP Echo Request(s) between endpoints. For each endpoint 1 ping packet will be sent and API shall wait for ping response to either be successful or timeout. The API wait timeout for each request is 300ms.";
   }
+  rpc SetProtocolState(SetProtocolStateRequest) returns (SetProtocolStateResponse) {
+    option (rpc_meta).description = "Sets all configured protocols to intended state";
```
apratimmukherjee, line 12107: started or stopped ?

control/control.yaml
```diff
@@ -124,4 +124,16 @@ components:
             Array of ping requests
           type: array
           items:
-            $ref: './ping.yaml#/components/schemas/Ping'
\ No newline at end of file
+            $ref: './ping.yaml#/components/schemas/Ping'
+
+    Protocol.State:
+      description: >-
+        Sets all configured protocols to intended state
```
apratimmukherjee, line 131: started or stopped state instead of intended state ?

artifacts/openapi.yaml
```diff
@@ -183,6 +183,26 @@ paths:
...
+      operationId: set_protocol_state
+      description: |-
+        Sets all configured protocols to intended state
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Protocol.State'
+      responses:
```
apratimmukherjee, line 199: Just a query , not a comment. In the generated .proto all the return types are of a single type called Response inside which there is an array of strings ( error strings or warning strings ) . What is the signficance of the Success/BadRequest and InternalServerError below. Is this hidden from user in the reply via .proto format and user sees these same responses as Response messages ?

apratimmukherjee (approved):
I have left some comments but none of them block commit. So approving the review.

artifacts/openapi.yaml
ashutshkumr: - `Success`: should contain warnings and any other useful data structures
- `BadRequest`: should at least contain a list of errors for validation errors (e.g. `pause` is not a correct state, `eth1` is not a valid device name, etc.)
- `InternalServerError`: should at least cotain a list of errors for actual errors internal to application (e.g. devices are all valid but starting protocols on some failed)

control/control.yaml
ashutshkumr: changed to mention `start` or `stop` state.

artifacts/otg.proto
ashutshkumr: changed to mention `start` or `stop` state.

api/api.yaml
ashutshkumr: changed to mention `start` or `stop` state.

ankur-sheth (approved):

ajbalogh:
@ByReaL has an immediate use case that requires starting a subset of protocols.

ashutshkumr:
@ajbalogh thanks for raising, we'll raise a separate PR for it. Do we have an ETA ?

ByReaL:
tried a test recently for a port monitor scenario, i have 3 ports in test on 2 of them protocols are up and on third one protocols are down on purpose, (bgp not the link) i need to start the traffic with bgp up on 2 ports and down on the third port but i cannot do that because is waiting for all protocols to be up.

how can we star traffic with protocols down ?

apratimmukherjee:
I guess set_protocol_state ( start) should return after starting the protocol and not wait for sessions to come up ?  Sessions up should be checked using BgpMetrics and when sessions are up and routes tx/rx ed fine as expected ( would be expected as sessions up on 2 ports in this case ) , then start traffic flows on relevant ports . [ This is the workflow with athena-g/ixia-c implementation ]

ajbalogh:
I think the description of the set_protocol_state should be enhanced with the description set out by @apratimmukherjee

PR #174, opened 2021-10-04
name attributes Isis Interface and Traffic Engineering

SuryyaKrJana:


device/isis/multipleinstances.yaml
```diff
@@ -2,6 +2,8 @@ components:
   schemas:
     Device.IsisMultiInstance:
       x-status: under-review
+      x-include:
+      - '../common/common.yaml#/components/schemas/Named.Object'
```
apratimmukherjee, line 6: Probably we can keep  it as a property container instead of actionable node.

result/isis.yaml
```diff
@@ -0,0 +1,187 @@
...
+            If the list is empty then all columns will be returned except for
+            any result_groups.
+            The name of the ISIS Router cannot be excluded.
+          type: array
+          items:
+            type: string
+            enum:
+              - sessions_state_not_started
+              - session_state_up
+              - session_state_down
```
apratimmukherjee, line 35: These 3 stats might be removable.

SuryyaKrJana:
ISIS Model bug fixing....

PR #175, opened 2021-10-06
Arp cache support

PrasenjitAdhikary:
[Please view the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/arp_cache_support/artifacts/openapi.yaml)

The proposed change has the support of retrieving ARP cache and Neighbor Discovery entries to make the same information available through gNMI interface (through the gNMI server) to a user.

The result set has been intentionally subdivided into two sub-groups, for resolved and unresolved entries, instead of adding one more column with resolved / unresolved status.  This approach has the benefit of not processing the entire result to know if there is any un-resolved entry and need a re-query before moving to the next step of the test. Instead, it is less overhead to look or zero entry in the unresolved group.

The column set for resolved / unresolved items is intentionally made identical, for easier processing, though the link-layer address column is always un-used for the unresolved group and can be removed.

Assuming the number of columns is limited (3 or 2 for the unresolved group), the column filter can also be removed from the request configuration (and also remove the link-layer address column from unresolved entries).

### DIFF
--- a/result/arp.yaml
+++ b/result/arp.yaml
@@ -24,31 +24,50 @@ components:
           description: >-
             The list of column names that the returned result set will contain.
             If the list is empty then all columns will be returned.
-            The column interface_name will always be present in the returned result set.
+            The column ipv4_name will always be present in the returned result set.
           type: array
           items:
             type: string
             enum:
-              - interface_name
-              - ip_address
+              - ipv4_name
+              - neighbor_ip
               - link_layer_address
+
+    Arp.Entries.Response:
+      description: >-
+        A container for ARP entries.
+      type: object
+      properties:
+        resolved:
+          description: >-
+            Contains the ARP entries where link-layer address of the neighbor is already known.
+          type: array
+          items:
+            $ref: './arp.yaml#/components/schemas/Arp.Entry'
+        unresolved:
+          description: >-
+            Contains the ARP entries where link-layer address learning of the neighbor is still pending.
+          type: array
+          items:
+            $ref: './arp.yaml#/components/schemas/Arp.Entry'
 
     Arp.Entry:
       description: >-
         ARP entry. A per neighbor information for IPv4 interfaces, learned through ARP.
       type: object
       properties:
-        interface_name:
+        ipv4_name:
           description: >-
-            The name of the interface.
+            The name of the IPv4 interface.
           type: string
-        ip_address:
+        neighbor_ip:
           description: >-
-            The IPv4 address of the neighbour node.
+            The IPv4 address of the neighbor node.
           type: string
           format: ipv4
         link_layer_address:
           description: >-
             The link-layer address of the neighbor node.
+            For an unresolved ARP entry, value will be either empty or set to default MAC address.
           type: string
           format: mac
--- a/result/arp.yaml
+++ b/result/arp.yaml
@@ -8,13 +8,13 @@ components:
   schemas:
     Arp.Entries.Request:
       description: >-
-        The request to retrieve ARP entries per IPv4 interface.
+        The request to retrieve ARP entries for IPv4 interface(s).
       type: object
       properties:
         ipv4_names:
           description: >-
             The names of IPv4 interfaces for which ARP entries will be retrieved.
-            If no interface names are specified then the result set will contain all ARP entries.
+            If no interface names are specified then the result set will contain all ARP entries for all available IPv4 interfaces.
           type: array
           items:
             type: string
@@ -22,9 +22,9 @@ components:
             - "/components/schemas/Device.Ipv4/properties/name"
         column_names:
           description: >-
-            The list of column names that the returned result set will contain.
-            If the list is empty then all columns will be returned.
-            The column ipv4_name will always be present in the returned result set.
+            Name of the columns (as list), that will present in the returned result set.
+            If the column names is empty then all columns will be returned.
+            The column ipv4_name (the name of the IPv4 interface) will always be present in the returned result set.
           type: array
           items:
             type: string
@@ -40,20 +40,20 @@ components:
       properties:
         resolved:
           description: >-
-            Contains the ARP entries where link-layer address of the neighbor is already known.
+            Contains the ARP entries for which link-layer address (MAC) of the neighbor is already known.
           type: array
           items:
             $ref: './arp.yaml#/components/schemas/Arp.Entry'
         unresolved:
           description: >-
-            Contains the ARP entries where link-layer address learning of the neighbor is still pending.
+            Contains the ARP entries for which link-layer address (MAC) learning of the neighbor is still pending.
           type: array
           items:
             $ref: './arp.yaml#/components/schemas/Arp.Entry'
 
     Arp.Entry:
       description: >-
-        ARP entry. A per neighbor information for IPv4 interfaces, learned through ARP.
+        ARP entry. For an IPv4 interface, a neighbor MAC information as learned through ARP.
       type: object
       properties:
         ipv4_name:
@@ -67,7 +67,7 @@ components:
           format: ipv4
         link_layer_address:
           description: >-
-            The link-layer address of the neighbor node.
+            The link-layer address (MAC) of the neighbor node.
             For an unresolved ARP entry, value will be either empty or set to default MAC address.
           type: string
           format: mac
--- a/result/arp.yaml
+++ b/result/arp.yaml
@@ -1,0 +1,54 @@
+openapi: 3.1.0
+
+info:
+  title: ARP table models
+  version: ^0.0.0
+
+components:
+  schemas:
+    Arp.Entries.Request:
+      description: >-
+        The request to retrieve ARP entries per IPv4 interface.
+      type: object
+      properties:
+        ipv4_names:
+          description: >-
+            The names of IPv4 interfaces for which ARP entries will be retrieved.
+            If no interface names are specified then the result set will contain all ARP entries.
+          type: array
+          items:
+            type: string
+          x-constraint:
+            - "/components/schemas/Device.Ipv4/properties/name"
+        column_names:
+          description: >-
+            The list of column names that the returned result set will contain.
+            If the list is empty then all columns will be returned.
+            The column interface_name will always be present in the returned result set.
+          type: array
+          items:
+            type: string
+            enum:
+              - interface_name
+              - ip_address
+              - link_layer_address
+
+    Arp.Entry:
+      description: >-
+        ARP entry. A per neighbor information for IPv4 interfaces, learned through ARP.
+      type: object
+      properties:
+        interface_name:
+          description: >-
+            The name of the interface.
+          type: string
+        ip_address:
+          description: >-
+            The IPv4 address of the neighbour node.
+          type: string
+          format: ipv4
+        link_layer_address:
+          description: >-
+            The link-layer address of the neighbor node.
+          type: string
+          format: mac
--- a/result/metrics.yaml
+++ b/result/metrics.yaml
@@ -12,6 +12,8 @@ components:
           - flow
           - bgpv4
           - bgpv6
+          - arp
+          - nd
           default: port
         port:
           $ref: './port.yaml#/components/schemas/Port.Metrics.Request'
@@ -21,6 +23,10 @@ components:
           $ref: './bgpv4.yaml#/components/schemas/Bgpv4.Metrics.Request'
         bgpv6:
           $ref: './bgpv6.yaml#/components/schemas/Bgpv6.Metrics.Request'
+        arp:
+          $ref: './arp.yaml#/components/schemas/Arp.Entries.Request'
+        nd:
+          $ref: './nd.yaml#/components/schemas/Nd.Entries.Request'
 
     Metrics.Response:
       description: >-
@@ -34,6 +40,8 @@ components:
           - port_metrics
           - bgpv4_metrics
           - bgpv6_metrics
+          - arp_entries
+          - nd_entries
           default: port_metrics
         port_metrics:
           type: array
@@ -51,3 +59,11 @@ components:
           type: array
           items:
             $ref: './bgpv6.yaml#/components/schemas/Bgpv6.Metric'
+        arp_entries:
+          type: array
+          items:
+            $ref: './arp.yaml#/components/schemas/Arp.Entry'
+        nd_entries:
+          type: array
+          items:
+            $ref: './nd.yaml#/components/schemas/Nd.Entry'
--- a/result/metrics.yaml
+++ b/result/metrics.yaml
@@ -61,9 +61,9 @@ components:
             $ref: './bgpv6.yaml#/components/schemas/Bgpv6.Metric'
         arp_entries:
           description: >-
-            Arp entries - for resolved & unresolved.
+            ARP entries - for resolved & unresolved ARP(s).
           $ref: './arp.yaml#/components/schemas/Arp.Entries.Response'
         nd_entries:
           description: >-
-            ND entries - for resolved & unresolved.
+            ND entries - for resolved & unresolved ND(s).
           $ref: './nd.yaml#/components/schemas/Nd.Entries.Response'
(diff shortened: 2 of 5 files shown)


result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+  schemas:
+    Arp.Entries.Request:
+      description: >-
+        The request to retrieve ARP entries per IPv4 interface.
+      type: object
+      properties:
+        ipv4_names:
+          description: >-
+            The names of IPv4 interfaces for which ARP entries will be retrieved.
+            If no interface names are specified then the result set will contain all ARP entries.
```
kingshukdev, line 17: If no interface names are specified then the result set will contain all ARP entries for all IPv4 interfaces.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            type: string
+          x-constraint:
+            - "/components/schemas/Device.Ipv4/properties/name"
+        column_names:
+          description: >-
+            The list of column names that the returned result set will contain.
```
kingshukdev, line 25: Name of the columns (as list), that will present in the returned result set.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+            If the list is empty then all columns will be returned.
```
kingshukdev, line 26: Which list are we referring here ?

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+            The column ipv4_name will always be present in the returned result set.
```
kingshukdev, line 27: The column ipv4_name (the name of the IPv4 interface) will always be present in the returned result set.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+              - link_layer_address
+
+    Arp.Entries.Response:
+      description: >-
+        A container for ARP entries.
+      type: object
+      properties:
+        resolved:
+          description: >-
+            Contains the ARP entries where link-layer address of the neighbor is already known.
```
kingshukdev, line 43: Contains the ARP entries for which link-layer address (MAC) of the neighbor is already known.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            $ref: './arp.yaml#/components/schemas/Arp.Entry'
+        unresolved:
+          description: >-
+            Contains the ARP entries where link-layer address learning of the neighbor is still pending.
```
kingshukdev, line 49: Contains the ARP entries for which link-layer address (MAC) learning of the neighbor is still pending.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            $ref: './arp.yaml#/components/schemas/Arp.Entry'
+
+    Arp.Entry:
+      description: >-
+        ARP entry. A per neighbor information for IPv4 interfaces, learned through ARP.
```
kingshukdev, line 56: For an IPv4 interface, a neighbor MAC information as learned though APR.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+            The name of the IPv4 interface.
+          type: string
+        neighbor_ip:
+          description: >-
+            The IPv4 address of the neighbor node.
+          type: string
+          format: ipv4
+        link_layer_address:
+          description: >-
+            The link-layer address of the neighbor node.
```
kingshukdev, line 70: The link-layer address (MAC) of the neighbor node.

result/metrics.yaml
```diff
@@ -51,3 +59,11 @@ components:
           type: array
           items:
             $ref: './bgpv6.yaml#/components/schemas/Bgpv6.Metric'
+        arp_entries:
+          description: >-
+            Arp entries - for resolved & unresolved.
```
kingshukdev, line 64: Arp entries - for resolved & unresolved ARP(s).

result/metrics.yaml
```diff
@@ -51,3 +59,11 @@ components:
...
+          $ref: './arp.yaml#/components/schemas/Arp.Entries.Response'
+        nd_entries:
+          description: >-
+            ND entries - for resolved & unresolved.
```
kingshukdev, line 68: ND entries - for resolved & unresolved ND(s).

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+info:
+  title: Neighbor Discovery table models
+  version: ^0.0.0
+
+components:
+  schemas:
+    Nd.Entries.Request:
+      description: >-
+        The request to retrieve Neighbor Discovery entries per IPv6 interface.
```
kingshukdev, line 11: The request to retrieve Neighbor Discovery entries for IPv6 interface(s).

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+      type: object
+      properties:
+        ipv6_names:
+          description: >-
+            The names of IPv6 interfaces for which ND entries will be retrieved.
+            If no interface names are specified then the result set will contain all ND entries.
```
kingshukdev, line 17: If no interface names are specified then the result set will contain all ND entries for all available interfaces.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            type: string
+          x-constraint:
+            - "/components/schemas/Device.Ipv6/properties/name"
+        column_names:
+          description: >-
+            The list of column names that the returned result set will contain.
```
kingshukdev, line 25: Same comments as give for ARP

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+              - link_layer_address
+
+    Nd.Entries.Response:
+      description: >-
+        A container for Neighbor Discovery entries.
+      type: object
+      properties:
+        resolved:
+          description: >-
+            Contains the Neighbor Discovery entries where link-layer address of the neighbor is already known.
```
kingshukdev, line 43: Contains the Neighbor Discovery entries where link-layer address (MAC) of the neighbor is already known.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            $ref: './nd.yaml#/components/schemas/Nd.Entry'
+        unresolved:
+          description: >-
+            Contains the Neighbor Discovery entries where link-layer address learning of the neighbor is still pending.
```
kingshukdev, line 49: Contains the Neighbor Discovery entries where link-layer address (MAC) learning of the neighbor is still pending.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+            The name of the IPv6 interface.
+          type: string
+        neighbor_ip:
+          description: >-
+            The IPv6 address of the neighbor node.
+          type: string
+          format: ipv6
+        link_layer_address:
+          description: >-
+            The link-layer address of the neighbor node.
```
kingshukdev, line 70: he link-layer address (MAC) of the neighbor node.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
+openapi: 3.1.0
+
+info:
+  title: ARP table models
+  version: ^0.0.0
+
+components:
+  schemas:
+    Arp.Entries.Request:
```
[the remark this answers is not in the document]
apratimmukherjee: ARP is probably better.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+      description: >-
+        The request to retrieve ARP entries per IPv4 interface.
```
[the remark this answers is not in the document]
apratimmukherjee: maybe 'for a set of IPv4 interfaces'

result/metrics.yaml
apratimmukherjee: Arp entries-> ARP entries

apratimmukherjee (approved):

result/arp.yaml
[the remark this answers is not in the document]
PrasenjitAdhikary: A similar convention is used for object names. example, Bgpv4.Metric

result/arp.yaml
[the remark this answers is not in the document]
PrasenjitAdhikary: Description updated.

result/arp.yaml
PrasenjitAdhikary: Updated.

result/arp.yaml
PrasenjitAdhikary: Updatred.

result/arp.yaml
PrasenjitAdhikary: Description updated.

result/metrics.yaml
PrasenjitAdhikary: Updated.

result/nd.yaml
PrasenjitAdhikary: Updated.

result/nd.yaml
PrasenjitAdhikary: Description updated.

result/nd.yaml
PrasenjitAdhikary: Description updated.

result/nd.yaml
PrasenjitAdhikary: Description updated.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+        unresolved:
+          description: >-
+            Contains the Neighbor Discovery entries where link-layer address learning of the neighbor is still pending.
+          type: array
+          items:
+            $ref: './nd.yaml#/components/schemas/Nd.Entry'
+
+    Nd.Entry:
+      description: >-
+        Neighbor Discovery entry. A per neighbor information for IPv6 interfaces, learned through NDP.
```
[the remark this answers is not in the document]
PrasenjitAdhikary: Description updated.

result/nd.yaml
PrasenjitAdhikary: Description updated.

result/arp.yaml
PrasenjitAdhikary: Description updated.

result/arp.yaml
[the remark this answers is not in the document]
ajbalogh: The modeling guide states the following: schema object names MUST be PascalCase (upper camel case)
https://github.com/open-traffic-generator/openapiart/blob/main/MODELGUIDE.md

ajbalogh (commented):
At what point should arp entries be available? Is there a specific control operation that triggers arp?

result/arp.yaml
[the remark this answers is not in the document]
apratimmukherjee: I meant for description text , not for object/attribute names. I believe for that we should use upper case for acronymns in description,  right. Agree for object name/attribute names, we need to use Pascal Case.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+        column_names:
+          description: >-
+            Name of the columns (as list), that will present in the returned result set.
+            If the column names is empty then all columns will be returned.
+            The column ipv6_name (the name of the IPv6 interface) will always be present in the returned result set.
+          type: array
+          items:
+            type: string
+            enum:
+              - ipv6_name
```
ashutshkumr, line 32: since `ipv6_name` is mandatory, it should be removed from list of enums in column names.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+info:
+  title: Neighbor Discovery table models
+  version: ^0.0.0
+
+components:
+  schemas:
+    Nd.Entries.Request:
+      description: >-
+        The request to retrieve Neighbor Discovery entries for IPv6 interface(s).
```
ashutshkumr, line 11: *retrieve ND (Neighbor Discovery) entries*

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+            enum:
+              - ipv4_name
+              - neighbor_ip
+              - link_layer_address
+
+    Arp.Entries.Response:
+      description: >-
+        A container for ARP entries.
+      type: object
+      properties:
```
ashutshkumr, line 40: Why don't we just add `resolved` as a column name where its value (true / false) for any given row should indicate the same info ?

IMHO, additional nesting of resolved vs unresolved is redundant here.

result/nd.yaml
```diff
@@ -0,0 +1,73 @@
...
+            enum:
+              - ipv6_name
+              - neighbor_ip
+              - link_layer_address
+
+    Nd.Entries.Response:
+      description: >-
+        A container for Neighbor Discovery entries.
+      type: object
+      properties:
```
ashutshkumr, line 40: Why don't we just add `resolved` as a column name where its value (true / false) for any given row should indicate the same info ?

IMHO, additional nesting of resolved vs unresolved is redundant here.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+  schemas:
+    Arp.Entries.Request:
+      description: >-
+        The request to retrieve ARP entries for IPv4 interface(s).
+      type: object
+      properties:
+        ipv4_names:
+          description: >-
+            The names of IPv4 interfaces for which ARP entries will be retrieved.
+            If no interface names are specified then the result set will contain all ARP entries for all available IPv4 interfaces.
```
ankur-sheth, line 17: will contain all ARP entries for all --> will contain ARP entries for all

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+          type: array
+          items:
+            type: string
+          x-constraint:
+            - "/components/schemas/Device.Ipv4/properties/name"
+        column_names:
+          description: >-
+            Name of the columns (as list), that will present in the returned result set.
+            If the column names is empty then all columns will be returned.
```
ankur-sheth, line 26: If column names is empty --> If column_names is empty

result/arp.yaml
ankur-sheth: > Why don't we just add `resolved` as a column name where its value (true / false) for any given row should indicate the same info ?
> 
> IMHO, additional nesting of resolved vs unresolved is redundant here.

What is the use-case for providing a list of unresolved entries in the first place?  What is that list expected to contain?  Just interface_name?

result/metrics.yaml
```diff
@@ -12,6 +12,8 @@ components:
           - flow
           - bgpv4
           - bgpv6
+          - arp
+          - nd
```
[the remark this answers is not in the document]
ankur-sheth: Agree, I'm not a big fan of including it in the results/metrics.

result/arp.yaml
```diff
@@ -0,0 +1,73 @@
...
+            The column ipv4_name (the name of the IPv4 interface) will always be present in the returned result set.
+          type: array
+          items:
+            type: string
+            enum:
+              - ipv4_name
+              - neighbor_ip
+              - link_layer_address
```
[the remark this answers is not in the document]
PrasenjitAdhikary: For consistency with OpenConfig term, where the column is referred to as link-layer-address instead of mac_address.

result/arp.yaml
ankur-sheth: I can't imagine a scenario where an end user (in a script) is going to go thru 1000 entries of an ARP table to see what resolved and what didn't.  And the way to determine if an IP is resolved or not, is just to go thru the resolved list to see if it is present there or not.  I don't know of any tool (or OS) that gives a list of unresolved IPs. 

I still think we are missing the use-case(s) around this feature.  

I don't think that ARP will be a test-case by itself.  Usually, there will be other protocols that will be running on top of IP and those will fail.  Typically, at that time, ARP is used as a diagnostic tool and that works well from a GUI app.  In case of an API driven environment, I'm not sure what value the ARP table would serve.

That said, I think a more important piece that _might_ be missing is the ability for the user to trigger ARP for a specific destination IP.  

Let's get the use-cases and all these questions will be automatically answered.

ashutshkumr:
@PrasenjitAdhikary kindly review and reopen if this needs to be alive.

PR #177, opened 2021-10-07
Reverting back LAG ethernet config

Rangababu-R:
@SuryyaKrJana / @apratimmukherjee,
The use case referring to is if Ethernet mac and vlan needs to configured on a LAG interface.
recently in sonic we noticed a use case with vlan being configured on a port-channel

please advice if this is a valid scenario?

### DIFF
--- a/lag/lag.yaml
+++ b/lag/lag.yaml
@@ -14,9 +14,9 @@ components:
     
     Lag.Port:
       description: >-
-        The container for a port and LAG protocol settings.
+        The container for a port's ethernet interface and LAG protocol settings
       type: object
-      required: [port_name, protocol]
+      required: [port_name, protocol, ethernet]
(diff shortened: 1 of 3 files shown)


lag/lag.yaml
```diff
@@ -26,6 +26,8 @@ components:
           - '/components/schemas/Port/properties/name'
         protocol:
           $ref: '#/components/schemas/Lag.Protocol'
+        ethernet:
+          $ref: '../device/ethernet.yaml#/components/schemas/Device.EthernetBase'
```
SuryyaKrJana, line 30: By mistaken it has been deleted.

apratimmukherjee (approved):
Yes, indeed looks like an unintended change. It should indeed be reverted.

PR #178, opened 2021-10-09
Adding bgp local preference

SuryyaKrJana:


device/bgp/bgproutepathattributes.yaml
```diff
@@ -18,4 +18,10 @@ components:
           enum:
           - igp
           - egp
-          default: igp
\ No newline at end of file
+          default: igp
+        local_preference:
```
apratimmukherjee, line 22: Might be good to also have a boolean field to exclude it if needed, e.g, include_local_preference , default true ; so that the default ibgp use-case works where this must be present but user has option to exclude from non-default ebgp usecase if desired . It could be made an optional required parameter but that makes is harder for users and also existing scripts need to have this explicitly added else iBGP sessions will flap, so we should definitely have the default value as done now.

apratimmukherjee (approved):

device/bgp/bgproutepathattributes.yaml
```diff
@@ -18,4 +18,10 @@ components:
...
+          description: >-
+            Local Preference attribute, for iBGP peers to select the path for the traffic leaving the AS.
+          type: integer
+          default: 100
```
apratimmukherjee, line 26: Maybe default of 0 is better since it aligns with current defaults and with default  in IxN today.

apratimmukherjee (approved):

device/bgp/bgproutepathattributes.yaml
ashutshkumr: Agreed, although I would advice against exposing another attribute. Maybe we can have local_preference as an object with one attribute called `value`. It will achieve similar purpose.

e.g. `obj.local_preference.value = 100`

device/bgp/bgproutepathattributes.yaml
ashutshkumr: what's wrong with 100 as a default value which is more universally recognized ?

PR #208, opened 2022-02-03
Bgp evpn

amrghosh:
This PR modelsBGP EVPN VXLAN support in Open Traffic Generator (OTG)
This is the proposal to add BGP EVPN VXLAN support to the OTG model as per earlier brainstorming and discussions.
[LINK](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/bgp_evpn/artifacts/openapi.yaml)

Demo Review meeting notes addressed : 
=====================================

1) The "ethernet_segments" name to be changed to "evpn_ethernet_segments". The hierarchy is "devices->bgp->ipv4_interfaces->peers->evpn_ethernet_segments"
2) AS number: Value. ==> formatted as tuple "AS number: Value".
2) In future the other VPNs(L2VPN/L3VPN/MVPN) would be included in a separate container under "peers". 
   So they would come under devices->bgp->ipv4_interfaces->peers->vrfs. 
   Unders "VRFs" we can have separate nodes for l2vrf, l3vrfs etc.
   
   It was specially noted that we should try to avoid adding a new parent to any node as that will break the existing scripts. 
   So special attention needs to be taken for that while modelling. Adding new childs are fine.
   
3) It was decided to let the user configure both evpn_vxlan as well as evpn_mpls on the same ethernet_segment index.
	So the modelling has been changed to evpn_ethernet_segments[] -> evis [] .
	Here each "evi" will have a choice to set either "evpn_vxlan" or "evpn_mpls" or anything else which can be added later to the choice.
	
4) "choice" is a special attribute which is being used specially for some UX. hence it was decided to have the options to select the evi types as a "choice" field
	and not as a generic enum field.

5) So each evi will have a --> {choice: "evpn_vxlan" , evpn_vxlan: {object} }

6) The container description should show the major EVPN Routes that it advertises. 

7) It was suggested to write a sample script for reference purpose and autogenerate the json config from that script.

Additional modifications : 
==============================
1) Removed active - from all the objects
2) Add evpn_mpls -- and check ahow to access -- 
		checked currently it has to be accessed by 
			evpn_vxlan_obj = p1_es_1.evis.evpn_vxlan()[-1]
			evpn_mpls_obj = p1_es_1.evis.evpn_mpls()[-1] 
			
		Currently exploring if it can be set by setting the "choice" field
		
3) Give a valid unicast mac address - required and default cannot be set together. 
Going ahead with the address as a required field rather than providing default.

4) Remove the required in "mac_addresses" container -- removed required from the "mac_addresses" container


amrghosh:
closing

PR #219, opened 2022-03-15
Support for device chaining

ANISH-GOTTAPU:
https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/device_chain/artifacts/openapi.yaml

Support for device chaining:

```import snappi

def test_device_chaining(api):
    api = snappi.api()
    config = api.config()

    # Lag
    lag1 = config.lags.lag(name="lag1")[-1]
    lp1 = lag1.ports.port(port_name=p2.name)[-1]
    lp1.protocol.static.lag_id = 1
    lp1.ethernet.name = "lp1-e"
    lp1.ethernet.mac = "aa:aa:aa:aa:aa:aa"

    d1, d2, d3 = config.devices.device(name='d1').device(name='d2').device(name='d3')

    # connected to physical port
    e1 = d1.ethernets.ethernet()[-1]
    e1.connection.port_name = p1.name
    e1.name = 'e1'
    e1.mac = '00:01:00:00:00:01'

    # connected to lag
    e2 = d2.ethernets.ethernet()[-1]
    e2.connection.lag_name = lag1.name
    e2.name = 'e2'
    e2.mac = '00:01:00:00:00:02'    

    # Create a VXLAN  device
    vxlan = d1.vxlans.add()[-1]
    vxlan.name = "VXLAN"
    
    # Connected to VXLAN 
    e3 = d3.ethernets.ethernet()[-1]
    e3.connection.vxlan_name = vxlan.name              
    e3.name = 'e3'
    e3.mac = '00:01:00:00:00:03'
```


ajbalogh (commented):
Since the connection has been delegated to an object with a choice why not distinguish between all types: port_name, lag_name, device_name? Each one of the individual leaf nodes can then contain a detailed description of what the connection does.

ANISH-GOTTAPU:
> Since the connection has been delegated to an object with a choice why not distinguish between all types: port_name, lag_name, device_name? Each one of the individual leaf nodes can then contain a detailed description of what the connection does.

yeah, I agree..updated as suggested

device/ethernet.yaml
```diff
@@ -1,58 +1,78 @@
...
+    Device.Connection:
+      description: Device connection to physical port, LAG port or another device.
+      type: object
+      properties:
+        choice:
+          description: port_name, lag_name or device_name
+          type: string
+          enum: [port_name, lag_name, device_name]
+        port_name:
+          description: Name of the port the device is connected.
```
ajbalogh, line 65: nit - Name of the port that the device is connected to.

device/ethernet.yaml
```diff
@@ -1,58 +1,78 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+        lag_name:
+          description: Name of the Lag the device is connected.
```
ajbalogh, line 70: nit - Name of the Lag the device is connected to.

device/ethernet.yaml
```diff
@@ -1,58 +1,78 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Lag/properties/name'
+        device_name:
+          description: Connected/Emulated on top of another device
```
ajbalogh, line 75: nit - Name of the device that acts as the parent of this device.  ( i'm not sure of this, we should solicit some more feedback on exactly what this description should be, it would be good to get feedback from @apratimmukherjee )

ajbalogh (changes requested):

Rangababu-R (approved):

device/ethernet.yaml
```diff
@@ -1,58 +1,84 @@
...
+        lag_name:
+          description: Name of the Lag the device is connected to.
+          type: string
+          x-constraint:
+          - '/components/schemas/Lag/properties/name'
+        device_name:
+          description: Name of the device that acts as the parent of this device.
+          type: string
+          x-constraint:
+          - 'components/schemas/Device/properties/name'
```
apratimmukherjee, line 84: Adding sub-set of my previous comments:
1. A device can have multiple L2 connectors. In reality we have to chain it to one of these L2 tunnels. Chaining to a device is hence ambiguous to actually implement. [ Unless we create an arbitrary implementation rule that it is always to be chained behind first L2 connector for example but that does not seem to scale ] . i.e. might want to connect to one of the many VxLAN interfaces in the parent DG or one of the L2GRE instances or one of the L2VPN instances. This does not seem to handle it. Other notes are mentioned in that comment. Chaining to a device using device name does not seem to suffice.

apratimmukherjee (changes requested):

ANISH-GOTTAPU:
@apratimmukherjee : We had internal discussion and would be going with both `port_name` and `connection` for time being and will add `deprecated` for `port_name` and will put a warning `port_name`  going to be deprecated from x release.

device/ethernet.yaml
apratimmukherjee: Note that my comment is not related to port_name. I am talking about connection to a device. That itself looks incorrect to me. This connector is trying to connect to a particular instance of one of many types of L2 tunnels e.g. VXLAN, L2VPN or L2GRE

ANISH-GOTTAPU: yes got it. I'm thinking to update the models in the below way to have backward compatibility and based on your point that connection to a device doesn't make much sense. Please check and let me know your comments

1) Keep `port_name` as it is and remove required for now. Will add deprecated and remove it in future release.
2) `connection` will have three choices.
                     a) port_name
                     b) lag_name
                     c) l2_connector
3) As of now l2_connector will have x-constraint as vxlan name, will add l2gre&l2vpn in future based on user's requirement.

**Note:** We have to keep either `port_name` or `connection` required, will make connection as required once port_name is deprecated.

Sample test
========
```
# Create a VXLAN  device
vxlan = d1.vxlans.add()[-1]
vxlan.name = "VXLAN"

# Connected to VXLAN L2 Connector
e3 = d3.ethernets.ethernet()[-1]
e3.connection.l2_connector = vxlan.name              
e3.name = 'e3'
e3.mac = '00:01:00:00:00:03'
```

ajbalogh: Since port_name and lag_name have been broken out lets extend choice with vxlan_name and keep with specific descriptions?

ajbalogh: I think port_name and connection cannot be required until port_name is dropped from the model.

device/ethernet.yaml
```diff
@@ -1,58 +1,84 @@
...
-        ipv6_addresses:
-          description: |-
-            List of global IPv6 addresses and their gateways.
-            The Link Local IPv6 address will be automatically generated.
-          type: array
-          items:
-            $ref: './ipv6.yaml#/components/schemas/Device.Ipv6'
-      
\ No newline at end of file
+components:
```
ankur-sheth, line 1: Why are all these initial lines coming up in the diff?  Makes it a bit hard to review.

device/ethernet.yaml
```diff
@@ -1,58 +1,84 @@
...
+      - '#/components/schemas/Device.EthernetBase'
+      description: >-
+        An Ethernet interface with IPv4 and IPv6 addresses.
+      type: object
+      properties:
+        connection: 
+          description: |-
+            Device connection to physical, LAG or another device.
+          $ref : '#/components/schemas/Device.Connection'
+        port_name:
```
ankur-sheth, line 41: I am assuming that this will be marked as deprecated and the description will be updated?

device/ethernet.yaml
ankur-sheth: > Since port_name and lag_name have been broken out lets extend choice with vxlan_name and keep with specific descriptions?

Agree with Andy.  Might be better to go with specific descriptions.

ankur-sheth (changes requested):

apratimmukherjee (approved):
Looks good to me.

device/ethernet.yaml
ANISH-GOTTAPU: updated

device/ethernet.yaml
ANISH-GOTTAPU: some issue after I have upgraded my vscode, please enable `Hide whitespace` to have exact diff

device/ethernet.yaml
ANISH-GOTTAPU: updated as suggested

device/ethernet.yaml
```diff
@@ -1,58 +1,85 @@
...
+        ipv6_addresses:
+          description: |-
+            List of global IPv6 addresses and their gateways.
+            The Link Local IPv6 address will be automatically generated.
+          type: array
+          items:
+            $ref: './ipv6.yaml#/components/schemas/Device.Ipv6'
+      
+    Device.Connection:
+      description: Device connection to physical port, LAG port or another device.
```
apratimmukherjee, line 64: I am not sure it is a device connection, rather possibly the connection of the Ethernet interface , to a port, LAG or VxLAN tunnel . ( Needs updating whenever we add more choices ) ,

device/ethernet.yaml
```diff
@@ -1,58 +1,85 @@
...
+      type: object
+      properties:
+        choice:
+          description: port_name, lag_name or device_name
```
apratimmukherjee, line 68: device_name -> vxlan_name

device/ethernet.yaml
```diff
@@ -1,58 +1,85 @@
...
+          type: string
+          enum: [port_name, lag_name, vxlan_name]
+        port_name:
+          description: Name of the port that the device is connected to.
```
apratimmukherjee, line 72: device -> Ethernet interface ;  is connected to -> is configured on ?

device/ethernet.yaml
```diff
@@ -1,58 +1,85 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+        lag_name:
+          description: Name of the Lag the device is connected to.
```
apratimmukherjee, line 77: device -> Ethernet interface ; connected to -> is configured on ?

device/ethernet.yaml
```diff
@@ -1,58 +1,85 @@
...
+          type: string
+          x-constraint:
+          - '/components/schemas/Lag/properties/name'
+        vxlan_name:
+          description: Name of the vxlan that acts as overlay for the device.
```
apratimmukherjee, line 82: Name of the VxLAN instance ( or VxLAN tunnel )  that this Ethernet interface is connected to.

device/ethernet.yaml
ANISH-GOTTAPU: updated

device/ethernet.yaml
ANISH-GOTTAPU: updated

device/ethernet.yaml
ANISH-GOTTAPU: updated

device/ethernet.yaml
ANISH-GOTTAPU: updated

PR #225, opened 2022-04-05
x-status-message extension

Rangababu-R:
Adding new extension to support more on x-status extension.
Ex:
```yaml
components:
    schemas:
        Some.Object:
            properties:
                some_property:
                    x-status:
                        status: deprecated
                        additional_information: some_object.some_property is deprecated instead use some_object.other_property
                other_property:
```


MODEL-GUIDE.md
```diff
@@ -135,6 +135,20 @@ The build script will enforce the following keyword conventions:
             enum: [current, deprecated, obsolete, under_review]
             default: current
   ```
+* `x-status-message`
```
ajbalogh, line 138: An alternate approach would be to change the x-status extension from an enum to an object to allow for more flexibility in the future.
```yaml
x-status:
  type: object
  properties:
    status:
      type: string
      enum: [current, deprecated, obsolete, under_review]
    additional_information:
      type: string
      description: Use this property to provide additional information about the status property.
```

ajbalogh (changes requested):

MODEL-GUIDE.md
Rangababu-R: yeah agreed, I will change accordingly

Rangababu-R: updated as per the comments @ajbalogh

ajbalogh (approved):

Issue #231, opened 2022-04-14
Add Isis v4 and v6 route range names to x-constraint of Route.State


SouravSinhaRoy:
Isis v4 and v6 route range names are missing in x-constraint of Route.State

PR #232, opened 2022-04-14
adding isis route ranges in x-constraint validation of route states

SouravSinhaRoy:
Add Isis v4 and v6 route range names to x-constraint of Route.State #231


apratimmukherjee (approved):
Looks good to me.

apratimmukherjee:
@ashutshkumr : The reason for doing this is 1) it is just something missing that we needed to fix , 2) I notice that the snappi team is probably working on a PR to validate x-constraints when this would cause trouble with ISIS tests . Hence, adding this fix in the OTG model . ( No effective change as of now , just two missing x-constraints added)

ashutshkumr (approved):
LGTM

Issue #228, opened 2022-04-12
Clarify set_transmit_state


shramroy:
### Issue
The set_transmit_state API does not provide enough guidance to implementers in the following scenarios:
- what should happen when a partial list of flows is given
- what should happen to metrics

PR #234, opened 2022-04-21
Add additional transmit state context

ajbalogh:
Addresses issue #228

### DIFF
--- a/control/control.yaml
+++ b/control/control.yaml
@@ -31,7 +31,7 @@ components:
 
     Transmit.State:
       description: >-
-        Control the transmit state of flows.
+        Flow transmit state.
       type: object
       required: [state]
       properties:
@@ -40,6 +40,9 @@ components:
             The names of flows to which the transmit state will be applied to.
             If the list of flow_names is empty or null the state will be applied
             to all configured flows.
+
+            If the list is not empty any flow that is not included in the list
+            of flow_names MUST be ignored and not included in the state change.
           type: array
           items:
             type: string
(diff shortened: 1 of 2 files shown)


control/control.yaml
```diff
@@ -40,6 +40,9 @@ components:
             The names of flows to which the transmit state will be applied to.
             If the list of flow_names is empty or null the state will be applied
             to all configured flows.
+
```
shramroy, line 43: @ajbalogh should it have the detailing about metrics?

shramroy (dismissed):

control/control.yaml
ajbalogh: @shramroy the first comment in the PR has uses cases and commit 2d61746 has more context information regarding clearing of metrics

control/control.yaml
```diff
@@ -48,6 +51,20 @@ components:
         state:
           description: >-
             The transmit state.
+
+            If the value of the state property is 'start' then all flows
+            that are included in the 'flow_names' property MUST have their
+            metric counters cleared.
+
+            If the value of the state property is 'pause' then all flows that
+            are included in the 'flow_names' property will be stopped but
```
ankur-sheth, line 60: Perhaps say "included in the  'flow_names' property will be paused but their metric..."?

Also, do we want to distinguish between resuming a paused flow and starting a stopping flow? ie, the available enums are "[start, stop, pause, resume]"?

ankur-sheth (commented):
Do we want to provide 'resume' as a way to distinguish between starting a stopped flow, vs resuming a paused flow?

control/control.yaml
ajbalogh: Instead of adding more options I am proposing to replace pause with an state option of 'start_with_no_clear'.  This simplifies the setting of state and explicitly calls out the not clearing of metric counters.

ajbalogh:
@ashutshkumr @ankur-sheth I've updated the PR to remove start_with_no_clear and include `pause`, `resume` and would like your review.

ankur-sheth (approved):
I like it :)

shramroy (approved):

shramroy:
It looks good. Can we merge it then?

ashutshkumr (approved):

PR #236, opened 2022-05-06
clear metrics api add

shramroy:


ajbalogh (changes requested):
This link provides additional information about `choice` https://github.com/open-traffic-generator/openapiart/blob/main/MODELGUIDE.md#keyword-limitations

control/control.yaml
```diff
@@ -51,6 +51,36 @@ components:
           type: string
           enum: [start, stop, pause]
 
+    Clear.Metrics:
+      description: >-
+        Clear metrics.
+      type: object
+      required: [metrics_choice]
```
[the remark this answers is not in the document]
ankur-sheth: Agree.  `metrics_choice` doesn't seem to be clear.  We can keep it as `choice` as Andy suggests.

control/control.yaml
```diff
@@ -51,6 +51,36 @@ components:
...
+      properties:
+        port_names:
+          description: >-
+            The names of port objects to.
```
ankur-sheth, line 62: comment is incomplete. Perhaps rephrase as "The names of port objects for which metrics will be cleared"?

control/control.yaml
```diff
@@ -51,6 +51,36 @@ components:
...
+            An empty or null list will clear metrics of all ports.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+        flow_names:
+          description: >-
+            The names of flow objects to.
```
ankur-sheth, line 71: Same comment as above.  Incomplete comment.

ankur-sheth (changes requested):

ankur-sheth:
Also a general comment, please provide pseudo-code and link to issue(s) in the PR header.

shramroy:
Way to use clear_metrics api - 
metrics = api.clear_metrics()
metrics.choice = 'ports'
metrics.ports = ["p1", "p2"]
api.clear_metrics(metrics)

control/control.yaml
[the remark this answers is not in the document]
shramroy: If I put choice then "all" has to be an defined object type as per model definition 
 # demonstrates how to model a choice
  components:
    schemas:
    Choice.Object:
      type: object
      properties:
        choice:
          type: string
          enum: [a, b, c]
          default: a
        a:
          $ref: '#/components/schemas/Choice.A'
        b:
          $ref: '#/components/schemas/Choice.B'
        c:
          $ref: '#/components/schemas/Choice.C'




But with the choice "all" I don't want to give any other input. 
If the "metrics_choice" term is not convincing I changed it to "options". Will it be okay?

control/control.yaml
shramroy: modified

control/control.yaml
shramroy: modified

ankur-sheth (approved):

ajbalogh (commented):
With the use cases detailed in PR (https://github.com/open-traffic-generator/models/pull/234) where is the need for this API?

Specifically what are the use case(s) where a user would want to clear metrics that isn't met by PR 234.

ankur-sheth:
> With the use cases detailed in PR (#234) where is the need for this API?
> 
> Specifically what are the use case(s) where a user would want to clear metrics that isn't met by PR 234.

Agree with @ajbalogh.  Perhaps we can park this till we really have a use-case for it?

ajbalogh:
Closing this PR until specific use cases dictate a need for it.

PR #237, opened 2022-05-12
Sample Ixload model script

waseembaig:
Ixload model sample script


test_http_config.py
```diff
@@ -0,0 +1,86 @@
+import snappi
+api = snappi.api(host="10.22.40.53", version="8.50.0.465")
```
ankur-sheth, line 2: Why do we need a 'version' field here?

ankur-sheth (changes requested):

ashutshkumr:
Closing on behalf of https://github.com/open-traffic-generator/models/issues/238

Issue #239, opened 2022-05-24
Need support for retrieving bgp/isis route information


ajbalogh:
# Issue
There is no API for retrieving learned routes from a BGP peer and/or ISIS router.

# Enhancement
Enhance the models and APIs to allow for the retrieval of learned route information from BGP peers and/or ISIS routers.

PR #240, opened 2022-05-24
BGP/ISIS route information

ajbalogh:
[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/issue_239/artifacts/openapi.yaml)

Addresses issue #239 

Use case: Need to be able to retrieve IPV4 Prefix information for all configured BGP peers.
```python
# get bgp peer ipv4 prefix state
states_request = api.states_request()
states_request.bgp_ipv4_prefixes.bgp_peer_names = []
states_response = api.get_states(states_request)
for bgp_ipv4_prefix in states_response.bgp_ipv4_prefixes:
    print(bgp_ipv4_prefix.bgp_peer_name)
    for prefix in bgp_ipv4_prefix.prefixes:
        print(
            f"{prefix.ipv4_address} {prefix.prefix_length} {prefix.ipv4_next_hop} {prefix.origin}"
        )
```


ajbalogh:
@PrasenjitAdhikary this PR is a draft at the moment as I would like your feedback on the proposed models for retrieving ipv4 prefix information from a BGP peer(s).  Once the model structure is agreed on then I will commit the remainder of the bgpv6/isisv4/isisv6 prefix models.

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,55 @@
+components:
+  schemas:
+    Bgpv4Prefix.State.Request:
+      description: >-
+        The request to retrieve IPv4 Prefix state of a BGPV4 Peer.
+      type: object
+      properties:
+        ethernet_names:
```
ankur-sheth, line 8: Was this supposed to be ethernet_names or bgp_peer_name or something like that?

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,55 @@
...
+        bgp_peer_name:
+          description: >-
+            The name of the BGPV4 Peer associated with the prefix state information.
+          type: string
+        prefixes:
+          type: array
+          items:
+            $ref: '#/components/schemas/Bgpv4Prefix.Prefix'
+
+    Bgpv4Prefix.Prefix:
```
SuryyaKrJana: In OTG Bgp.V4RouteRange, today we already support 'Path_id', 'AS_Path', 'Community', etc. If first two can be also be supported in this?

SuryyaKrJana (changes requested):

result/bgpv4prefix.yaml
```diff
@@ -2,14 +2,14 @@ components:
   schemas:
     Bgpv4Prefix.State.Request:
       description: >-
-        The request to retrieve IPv4 Prefix state of a BGPV4 Peer.
+        The request to retrieve IPV4 Prefix state of a BGPV4 Peer.
```
ankur-sheth, line 5: Are we consistently using IPV4 or IPv4?  To me the latter seems more appropriate.  Don't care too much either ways, as long as we are consistent throughout the model.

ankur-sheth (approved):

PrasenjitAdhikary (approved):

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,55 @@
...
+          description: >-
+            The length of the IPv4 prefix.
+          type: integer
+        ipv4_next_hop:
+          type: string
+          format: ipv4
+        ipv6_next_hop:
+          type: string
+          format: ipv6
+        origin:
```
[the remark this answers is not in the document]
ajbalogh: @PrasenjitAdhikary thanks for reviewing the model.  Additional properties can be added as long as the initial model is flexible enough to accommodate them which it appears to be.  Using this as a template I will proceed with adding bgpv6 and isis models.

result/bgpv4prefix.yaml
SuryyaKrJana: Something in my mind and how it will look if a test client just will need to request through the OTG API “Get BGP IPv4 Unicast Routes”, “Get BGP IPv6 Unicast Routes”, “Get ISIS IPv4 Prefix Routes” or “Get ISIS IPv6 Prefix Routes”.
In response it will get “properties” of string key-values pairs array.
One such row entry of the array can be “{key-1:value-1, key-2:value-2, …,key-n:value-n}”.
For an example, for “Get BGP IPv4 Unicast Routes”:
“[IPv4 Prefix:202.1.1.0, IPv4 Next Hop:200:0:0:1, Path ID:10, AS Path:{1}]”

Then we don’t care about numbers fields to be added in future.

SuryyaKrJana: Something in my mind and how it will look if a test client just will need to request through the OTG API “Get BGP IPv4 Unicast Routes”, “Get BGP IPv6 Unicast Routes”, “Get ISIS IPv4 Prefix Routes” or “Get ISIS IPv6 Prefix Routes”.
In response it will get “properties” of string key-values pairs array.
One such row entry of the array can be “{key-1:value-1, key-2:value-2, …,key-n:value-n}”.
For an example, for “Get BGP IPv4 Unicast Routes”:
“[IPv4 Prefix:202.1.1.0, IPv4 Next Hop:200:0:0:1, Path ID:10, AS Path:{1}]”

Then we don’t care about numbers fields to be added in future.

SuryyaKrJana (approved):

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,73 @@
...
+        ipv4_address:
+          description: >-
+            The IPv4 prefix.
+          type: string
+          format: ipv4
+        prefix_length:
+          description: >-
+            The length of the IPv4 prefix.
+          type: integer
+        ipv4_next_hop:
```
apratimmukherjee, line 48: Does this need a next hop type ( ipv4/ipv6 ) --> how is it in the gnmi model for ate/dut that is currently defined.

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,73 @@
...
+        ipv6_next_hop:
+          description: >-
+            The IPv6 address of the egress interface.
+          type: string
+          format: ipv6
+        origin:
+          description: >-
+            The origin of the prefix.
+          type: string
+          enum: [igp, egp]
```
apratimmukherjee, line 62: There should be a third time , 'incomplete'

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,73 @@
...
+        path_id:
```
apratimmukherjee, line 63: path_id and even as_path might be optional. Is it expected that optional attribs will just not be present in the response ? Should this be added to documentation of such attributes. ( most will be optional except prefix/next-hop)

result/bgpv4prefix.yaml
ajbalogh: this is otg openapi specific not related to ate/dut gnmi

result/bgpv4prefix.yaml
ajbalogh: openapi attributes are by default all optional unless marked as required in the object

result/bgpv4prefix.yaml
```diff
@@ -0,0 +1,73 @@
...
+          description: >-
+            The path id.
+          type: integer
+        as_path:
```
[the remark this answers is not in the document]
ajbalogh: Since there is no specific requirement/test case i will remove these.

PrasenjitAdhikary (approved):
LGTM

PR #243, opened 2022-06-06
Dev lacp update

PrasenjitAdhikary:
[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-lacp-update/artifacts/openapi.yaml)

Ref doc for the changes: [LACP-OTGProposalV2.docx](https://keysighttech.sharepoint.com/:w:/r/sites/Athena-G/Shared%20Documents/Architecture/LAG/LACP-OTGProposalV2.docx?d=w5cf78404d195459ebf17d4518f56faa0&csf=1&web=1&e=KevsOC)


result/lacplagmember.yaml
```diff
@@ -0,0 +1,200 @@
...
+          description: >-
+            The name of a LAG (ports group) configured with LACP.
+          type: string
+        lag_member_name:
+          description: >-
+            The name of a LAG member (port) configured with LACP.
+          type: string
+        lacp_in_pkts:
+          description: >-
+            Number of LACPDUs received.
```
SuryyaKrJana, line 62: A space/hyphen between two words in LACPPDU may easy readable. The same has been occurred other placed too.

SuryyaKrJana (changes requested):

result/lacplagmember.yaml
PrasenjitAdhikary: Attribute description are based on OpenConfig model. It is one word - LACPDU, instead of LACP PDU.
https://github.com/openconfig/public/blob/master/release/models/lacp/openconfig-lacp.yang

lag/lag.yaml
```diff
@@ -114,20 +187,25 @@ components:
...
 
             Passive indicates the port's preference for not transmitting 
-            LACPDUs unless its partner's control is Active
+            LACPDUs unless its partner's control is Active.
 
             Active indicates the port's preference to participate in the 
-            protocol regardless of the partner's control value
+            protocol regardless of the partner's control value.
+            
+            Per port configuration is depricated, instead should be configured per LAG.
```
apratimmukherjee, line 207: depricated -> deprecated.

lag/lag.yaml
```diff
@@ -114,20 +187,25 @@ components:
             auto: 0
           description: >-
             This timer is used to detect whether received protocol information
-            has expired
+            has expired.
+            Per port configuration is depricated, instead should be configured per LAG.
```
apratimmukherjee, line 191: depricated->deprecated. Please check rest for similar typo.

apratimmukherjee (approved):
Some minor typos need to be fixed. Rest looks okay.

api/api.yaml
```diff
@@ -191,6 +191,26 @@ paths:
           $ref: '../result/request.yaml#/components/responses/BadRequest'
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
+  
+  /control/protocols/lacp:
```
ajbalogh, line 195: This approach has the potential of significantly increasing the size of the api surface.  A single API with the ability to extend the model should be proposed.

lag/lag.yaml
```diff
@@ -29,12 +41,26 @@ components:
...
+    Lag.Ports.Protocol:
+      type: object
+      properties:
+        choice:
+          description: >-
+            The type of controlling protocol for the LAG (ports group).
+          type: string
+          enum: [lacp, static]
+          default: lacp
+        lacp:
```
ajbalogh, line 53: Why is lacp being duplicated between ports and protocols?

ajbalogh (changes requested):

lag/lag.yaml
PrasenjitAdhikary: In OpenConfig or DUT, few LACP configuration properties are per LAG, not per port. In the existing model, such per LAG configuration is not allowed. Instead to achieve the same goal, the same set of configurations needs to be duplicated for each LAG member port. These could lead to inconsistency and undefined behavior. In IxNetwork, the modeling was done on per-port basis, as there was no option earlier to configure anything at Port Group level. Now as we are modeling in OTG, we can continue with the legacy model or we have the option to correct the model before it goes to the customer, the way it is used by the customer for other devices. So, few properties are duplicated to “ports” from “port” and properties at “port” level are marked deprecated.

api/api.yaml
PrasenjitAdhikary: Agree. Moved under a generic API

lag/lag.yaml
ajbalogh: If LACP configuration is done at the LAG level does this mean configuration will never be done at the port level? Or is this just a convenience option?

control/action.yaml
```diff
@@ -0,0 +1,44 @@
+openapi: 3.1.0
+
+info:
+  title: Action models
+  description: >-
+    All Detail Action schemas
```
winstonliu1111, line 6: @ajbalogh @PrasenjitAdhikary probably needs a more elaborate description here, it's not clear what/why/how it is applied.

control/control.yaml
```diff
@@ -165,4 +165,16 @@ components:
           description: >-
             Protocol specific states
           type: string
-          enum: [start, stop]
\ No newline at end of file
+          enum: [start, stop]
+
+    Action.Request:
+      description: >-
+        Request to traffic generator for top level choice of protocol/traffic/action-group having the detail action to perform
```
winstonliu1111, line 172: rephrase the description to make this easier to understand.

api/api.yaml
```diff
@@ -191,6 +191,26 @@ paths:
           $ref: '../result/request.yaml#/components/responses/BadRequest'
         '500':
           $ref: '../result/request.yaml#/components/responses/InternalServerError'
+  
+  /control/actions:
```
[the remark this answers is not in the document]
winstonliu1111: another related question, will "flap" a link or a protocol session be considered an "action"?

lag/lag.yaml
```diff
@@ -4,13 +4,25 @@ components:
...
+        protocol:
+          $ref: '#/components/schemas/Lag.Ports.Protocol'
+        min_links:
+          description: >-
+            Specifies the mininum number of member interfaces that must be active
+            for the aggregate interface to be available
+          type: integer
+          format: uint16
+          minimum: 0
+          maximum: 65535
```
winstonliu1111, line 24: protocol spec aside, is this testable by us in real life?  what is a reasonable max?

lag/lag.yaml
```diff
@@ -4,13 +4,25 @@ components:
...
+          default: 0
```
winstonliu1111, line 25: is this a reasonable default?

lag/lag.yaml
```diff
@@ -56,9 +82,52 @@ components:
           maximum: 65535
```
winstonliu1111, line 82: is this reasonable max/default?

result/lacplagmember.yaml
```diff
@@ -0,0 +1,148 @@
...
+        timeout:
+          description: >-
+            The timeout type (short or long) used by the participant.
+          type: string
+          enum: [short, long]
+        synchronization:
+          description: >-
+            Indicates whether the participant is in-sync or out-of-sync.
+          type: string
+          enum: [in_sync, out_sync]
```
winstonliu1111, line 101: out_sync -> out_of_sync

winstonliu1111 (changes requested):
can we get an example snappi based code snippet to show how it looks like? I think andy asked for this earlier as well.  want to get a feel for the ux by looking at the snippet.

api/api.yaml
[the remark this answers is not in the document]
PrasenjitAdhikary: (a) Similar triggers/actions are possible for BGP, ISIS, and more protocols in the future. Note - RouteWithdraw is one such action that is already added as a separate API. For example, in BGP peers only, we could have break/resume TCP sessions, Pause/Resume Keep alive, and do graceful restart are some example. Other protocols can have similar events at different object hierarchies.
(b) Another event for LACP, Member port up/down. Probably we will add this event after more discussion.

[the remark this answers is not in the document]
PrasenjitAdhikary: Currently, we have no direct way to bring down (flap) LACP link. This proposed new event would bring down the LACP link.

lag/lag.yaml
PrasenjitAdhikary: // Existing way of configuring LAG & LAG-member ports
lag1 := gosnappi.NewLag().SetName("lag1")
lag1.Ports().Add().SetPortName("lag1-p1").Protocol().Lacp().
		SetActorSystemId("01:01:01:01:01:01").SetActorSystemPriority(1).SetLacpduTimeout(1).SetActorPortNumber(1)
lag1.Ports().Add().SetPortName("lag1-p2").Protocol().Lacp().
		SetActorSystemId("01:01:01:01:01:01").SetActorSystemPriority(1).SetLacpduTimeout(1).SetActorPortNumber(2)
lag1.Ports().Add().SetPortName("lag1-p3").Protocol().Lacp().
		SetActorSystemId("01:01:01:01:01:01").SetActorSystemPriority(1).SetLacpduTimeout(1).SetActorPortNumber(3)

// New Proposal
lag1 = gosnappi.NewLag().SetName("lag1").Protocol().Lacp().SetActorSystemId("01:01:01:01:01:01").SetActorSystemPriority(1).SetLacpduTimeout(1)
lag1.Ports().Add().SetPortName("lag1-p1").Protocol().Lacp().SetActorPortNumber(1)
lag1.Ports().Add().SetPortName("lag1-p2").Protocol().Lacp().SetActorPortNumber(2)
lag1.Ports().Add().SetPortName("lag1-p3").Protocol().Lacp().SetActorPortNumber(3)

To avoid breaking old scripts, existing ways of configuring LAG-member ports are allowed (ideally we should make these lag-member attributes obsolete, instead of deprecate, but these are used by scripts for IxNetwork), but the recommendation is to configure LAG specific LACP parameters (i.e. ActorSystemId, ActorSystemPriority, LacpduTimeout & ActorActivity) at the LAG level, instead of repeating the same set of values for each LAG-member. And each LAG-member can not work with a different set of values, instead, any discrepancy of these parameters across LAG members are actually an invalid configuration.

.github/workflows/workflow.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
       - name: Commit openapi.yaml and otg.proto
         id: commit_generated
-        if: github.event_name == 'pull_request'
+        # if: github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
PrasenjitAdhikary: This was done intentionally - otherwise, artifacts were not getting generated on code submission. Ashutosh will verify with snappi team if this new option should be removed permanently.

result/lacplagmember.yaml
PrasenjitAdhikary: Based on OpenApi option flags. If we make it "out_of_sync", it can be odd for the user.
https://github.com/openconfig/public/blob/master/release/models/lacp/openconfig-lacp.yang

lag/lag.yaml
PrasenjitAdhikary: will update default to 1.

lag/lag.yaml
PrasenjitAdhikary: There will be validation in the implementation,  based on presently supported max. For example, Athena would validate for max value of 4 for now, which will increase to 8 or more gradually. In a different implementation (example in IxNetwork) - the limit can be different. So, should we change the max limit to model, or set it to max in the model and update the validation limit in the implementation, which can change from time to time.

lag/lag.yaml
PrasenjitAdhikary: This one is an arbitrary identifier - could be any value in the range, even for one lag.

PrasenjitAdhikary: And this attribute is moved from per Lag-member level to per Lag. No change in the existing limit.

api/api.yaml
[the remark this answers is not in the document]
winstonliu1111: @ajbalogh are we good with this action node given by Prasenjit?  what do you recommend with actions that have already gotten through via other api endpoints (Prasenjit brought up the routewithdraw example)

lag/lag.yaml
winstonliu1111: will there ever be an implementation that implements a LAG with 65535 members?  If not, what is a sensible default?  16, 32, 64, 128?  What is it?

ajbalogh: I agree with a small reasonable default - it can always be expanded in the future.

result/lacplagmember.yaml
winstonliu1111: Got it, OpenConfig calls it OUT_SYNC.  thanks.

lag/lag.yaml
winstonliu1111: I see the difference now.  thanks.

winstonliu1111 (changes requested):
still a few threads requiring resolution.

api/api.yaml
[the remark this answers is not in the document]
ajbalogh: To limit the spread of the api surface this action api should consolidate other single use protocol based apis  such as `set_protocol_state`, `set_route_state`, `send_ping` under one new api `set_protocol_states`. The new api should present a choice at the top level with choices for `protocol_state`, `route_state`, `lacp_pdu_state`. The choice allows for expansion of setting protocol states without increasing the top level api surface. The individual use protocol apis should be deprecated with the description for each pointing users to the new consolidated `set_protocol_states` api.

control/action.yaml
ajbalogh: The info block is redundant and not required, there is only one info block object which is found in the top level info.yaml file.

lag/lag.yaml
PrasenjitAdhikary: I have reduced the max value to 32 for now. But this field value is not the actual limit of LAG members. It just says how many members to be 'UP' to consider LAG is also in 'UP' state. The actual limit of the lag member should be in "ports" property, which is of type 'array' of 'port' object. @ajbalogh , can we add a limit here?

lag/lag.yaml
PrasenjitAdhikary: default updated to 1

lag/lag.yaml
PrasenjitAdhikary: After discussing with Andy, per port protocol attributes that are now available per LAG, have been removed from per port level. Now we have just one way of configuration. The following properties are removed from per LAG-member (i.e. port):
- protocol choice
- lacp:
       actor_system_id, actor_system_priority, lacpdu_timeout, actor_activity
- static:
       lag_id

This is a breaking change.

control/control.yaml
PrasenjitAdhikary: I have modified the description. But looks like this description block is not visible in ReDoc model.

api/api.yaml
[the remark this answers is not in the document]
PrasenjitAdhikary: Agree. Do the consolidation and deprecation in a different PR, in a subsequent sprint, This will help to start work on LACP implementation immediately and have the time to gradually deprecate/migrate to consolidate API.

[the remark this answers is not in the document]
winstonliu1111: @PrasenjitAdhikary  the issue linked to this consolidation/deprecation should be in the next sprint. we need to clean this up before wider use.

lag/lag.yaml
ajbalogh: @PrasenjitAdhikary you can add maxItems and minItems to the ports array but an implementation MUST return a run time error if it is unable to meet them.  Is minItems=1 and maxItems=8 appropriate?

api/api.yaml
```diff
@@ -191,6 +191,26 @@ paths:
...
+    post:
+      tags: ['Control']
+      operationId: set_action
```
ajbalogh, line 198: should be set_device_state to align with the config device model

api/api.yaml
```diff
@@ -191,6 +191,26 @@ paths:
...
+      description: >-
+        Set specific action/trigger on configuration resources on the traffic generator.
```
ajbalogh, line 200: This should be limited to specific state/actions on device configuration resources

control/action.yaml
```diff
@@ -0,0 +1,44 @@
+openapi: 3.1.0
+
+info:
```
ajbalogh, line 3: remove this object from this file as it is not required and will be overwritten

control/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+  title: Action models
+  description: >-
+    All Detail Action schemas
+  version: ^0.0.0
+
+components:
+  schemas:
+    Lacp.Action.Request:
```
ajbalogh, line 11: this is lacp specific and should be renamed as device state at the top level. choices can then reflect specific states.

control/action.yaml
```diff
@@ -0,0 +1,44 @@
+openapi: 3.1.0
+
+info:
+  title: Action models
```
ajbalogh, line 4: Lets not use action models - lets start this off as DeviceState and with this lacp state as the first choice.  This can then be expanded in subsequent PRs.

ajbalogh (changes requested):

lag/lag.yaml
PrasenjitAdhikary: Ports limit minItems=1 and maxItems=32 looks reasonable. At present protocol, engine supports 4 ports. So, implementation will through validation error if the ports count is more than 4. We will enhance the validation limit as and when support for more ports is added in Athena / Protocol engine.

api/api.yaml
PrasenjitAdhikary: modified set_action --> set_device_state

api/api.yaml
PrasenjitAdhikary: done

control/action.yaml
PrasenjitAdhikary: done

control/action.yaml
PrasenjitAdhikary: done.

control/action.yaml
PrasenjitAdhikary: done.

control/control.yaml
```diff
@@ -165,4 +165,39 @@ components:
...
+      properties:
+        port_names:
+          description: >-
+            The names of LAG member port objects to control.
+            An empty or null list will control all LAG member port objects.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
```
ajbalogh, line 196: this should be /components/schemas/Lag.Port/properties/name

ajbalogh (changes requested):

winstonliu1111 (commented):
@ajbalogh @PrasenjitAdhikary does it make sense to group all the lacp device state and parameters under a devices/lacp.yaml file that we can reference?  seems like lacp we're scattering them into different .yml files as they are referenced/used, unlike other protocols/concepts.

control/control.yaml
PrasenjitAdhikary: Done

PrasenjitAdhikary:
In the current hierarchy, only the device configuration and its children are under 'device' folder. All Metrics and States are grouped in 'result' folder. All control events are grouped in 'control' folder. Matching this, I have moved lacp part of the control event into a separate file under 'control' folder.

lag/lag.yaml
```diff
@@ -29,20 +43,26 @@ components:
...
           type: string
           enum: [lacp, static]
           default: lacp
         lacp:
-          $ref: '#/components/schemas/Lag.Lacp'
+          $ref: '#/components/schemas/Lag.Ports.Lacp'
         static:
           $ref: '#/components/schemas/Lag.Static'
 
+    Lag.Protocol:
```
ajbalogh, line 60: Since the lag.port.protocol is now only a container for the lacp container should this be removed and the lacp container moved up to lag.port.lacp with the lacp container being optional?

PrasenjitAdhikary: Done. The only concern, if we need to add similar protocols in the future (say MACsec / microBFD), do we need this protocol container?

ajbalogh (approved):
pending backward compatibility break - LGTM

lag/lag.yaml
winstonliu1111: @apratimmukherjee do we need this container?  wouldn't macsec/microbfd be things run over the LAG, but not to signal the LAG?

lag/lag.yaml
```diff
@@ -4,28 +4,42 @@ components:
...
+          maxItems: 32
+
+        protocol:
+          $ref: '#/components/schemas/Lag.Protocol'
+        min_links:
+          description: >-
+            Specifies the mininum number of member interfaces that must be active
+            for the aggregate interface to be available
+          type: integer
+          format: uint16
```
ANISH-GOTTAPU, line 24: `uint16` support is not there for generated sdks. Only int32&int64 are supported

ashutshkumr:
Aren't `SetProtocolState` and `SetDeviceState` similar ?
i.e. if we had a choice called `all` in `SetDeviceState`, it would eventually be equivalent to `SetProtocolState`

ANISH-GOTTAPU (approved):

PR #244, opened 2022-06-15
Enhance BGP prefix state API

ajbalogh:
[View the proposed changes using the redocly API viewer](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/bgp_prefixes/artifacts/openapi.yaml)

`Issue`
The BGP prefix state API has the following problems:
- the response does not allow for distinguishing between different types of prefix such as unicast ipv4 and unicast ipv6
- the request does not allow for filtering the response by prefix filter criteria

`Fix`
- Extend the request to allow for prefix filter criteria
- Enhance the response to provide specific prefix containers

### Select all prefixes for all peers
```yaml
bgp_prefixes:
  bgp_peer_names: []
  prefix_filters: []
  ipv4_unicast_filters: []
  ipv6_unicast_filters: []
```

### Select only ipv4_unicast prefixes for all peers
```yaml
bgp_prefixes:
  bgp_peer_names: []
  prefix_filters: [ipv4_unicast]
  ipv4_unicast_filters: []
  ipv6_unicast_filters: []
```


result/bgpprefix.yaml
```diff
@@ -16,6 +16,53 @@ components:
           x-constraint:
             - "/components/schemas/Bgp.v4Peer/properties/name"
             - "/components/schemas/Bgp.v6Peer/properties/name"
+        ipv4_unicast_filters:
```
PrasenjitAdhikary, line 19: How the user will specify that he is interested only in ipv4_unicast or ipv6_unicast prefixes? If "ipv4_unicast_filters" is empty or null, what will be the interpretation - return all or nothing? We need to cover the following:
(a) return all prefixes ==> Assuming no filter should support this scenario
(b) return all ipv4_unicast prefixes only or ipv6_unicast prefixes only ==> if this is supported?
(c) return specific ipv4/v6 unicast prefix ==> this is supported with the filter.

ajbalogh: https://github.com/open-traffic-generator/models/pull/244/commits/24f9d8ecc5963bef4c07edd6c4b4cc022a7837dd

PrasenjitAdhikary (approved):
Looks good to me.

PR #247, opened 2022-06-20
 Support for conditional storage of BGP IPv4 and IPv6 Unicast Learned…

apratimmukherjee:
Two new optional flags added in BGP peer to control whether Unicast IPv4 /IPv6 routes should be stored or not. ( Default is : dont store routes )


device/bgp/bgplearnedinfofilter.yaml
```diff
@@ -0,0 +1,18 @@
+components:
+  schemas:
+    Bgp.LearnedInformationFilter:
+      x-status: under-review
+      description: >-
+        Configuration for controlling storage of BGP learned information recieved from the DUT.
```
ashutshkumr, line 6: shouldn't it be `peer` instead of `DUT` ?

ashutshkumr (approved):

PrasenjitAdhikary (approved):
Looks good to me.

device/bgp/bgplearnedinfofilter.yaml
apratimmukherjee: Updated to peer from DUT .

SuryyaKrJana (approved):
Look good

PR #248, opened 2022-06-29
Dev dev api consolidate

PrasenjitAdhikary:
Deprecate all protocols related control API and consolidate them under set_device_state

document link: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-dev-api-consolidate/artifacts/openapi.yaml&nocors


control/control.yaml
```diff
@@ -176,5 +176,14 @@ components:
           type: string
           enum:
           - lacp_member_state
```
PrasenjitAdhikary, line 178: If changing the enum sequence does not break compatibility, I would prefer to keep mostly used events first - with the following sequence: 
      protocol_state
      route_state
      lacp_member_state
      send_ping

control/control.yaml
```diff
@@ -176,5 +176,14 @@ components:
...
+          - protocol_state
+          - route_state
+          - send_ping
```
ANISH-GOTTAPU, line 181: I think `send_ping` under `set_protocol_state` looks out of place. If you have given lot of thought and this should be part of `set_protocol_state` then its fine.

apratimmukherjee (approved):
Looks good to me.

control/control.yaml
ashutshkumr: I agree with Anish here. send_ping seems more like an action than a mere state change. Similar concern for advertising/withdrawing routes.

apratimmukherjee: All of these are indeed actions on some endpoint in the device tree . Implicitly they might or might not change state .
lacp_member_state -> down : Sends certain LACP PDU on the wire and then stops sending PDUs , up: Restarts sending regular LACP PDUs, Both of these might change FSM state based on what DUT does as a result, but not anything in config. 
set_protocol_state -> Starts sending BGP , ISIS, LACP, RSVP-TE PDUs on the wire, FSM state will change based on what connected device will do. Does not change config. 
set_route_state -> Sends ISIS/BGP withdraw routes / advertise routes PDU on the wire. Does not change config. 
send_ping -> Sends ping on wire. Might result in refresh of ARP cache on DUT/Ixia side. Does not change state or config. 
So , for me , all of them are actions ( that would either send single PDU or periodic PDUs or stop sending PDUs) that might or might not result in FSM state change on DUT or test port or both.
But the idea to consolidate them is based more on the fact that all these actions are on endpoints inside the Device tree . So from that point of view ping or traceroute also look to me to be actions on endpoints inside Device Tree , like advertise/withdraw routes or start/stop protocols or bring down/bring up LACP on specified LAG member ports. 
One could argue whether set_device_state could be also named as perform_device_action or set_device_action etc. ; however  the intention seems to me to bring all actions under devices in a single API to keep a single top-level API for all such actions, regardless of whether they change 'state' or not.

ashutshkumr: I guess what I'm referring to is the difference between actions that control `session state` and actions that do not.
e.g.
- advertising / withdrawing routes does not affect the session
- sending ping does not affect the session

api/api.yaml
```diff
@@ -130,6 +130,7 @@ paths:
       operationId: set_route_state
       description: >-
         Updates the state of configuration resources on the traffic generator.
+      x-status: deprecated
```
Rangababu-R, line 133: x-status has changed to type object. please change as per the model guide.
https://github.com/open-traffic-generator/models/blob/master/MODEL-GUIDE.md

Rangababu-R (changes requested):

Rangababu-R (approved):

PR #257, opened 2022-07-31
Need to maintain consistent .proto message field numbers

alakendu:
This should address https://github.com/open-traffic-generator/openapiart/issues/343


winstonliu1111 (commented):
consider refine this to reflect end user modeler's workflow/lifecycle.

group by 1) adding new properties/objects. 2) deprecating properties/objects 3) removing properties/objects

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
...
+```yaml
+components:
+  schemas:  
+    Prefix.Config:
+      properties:
+        ieee_802_1qbb:
+          type: boolean
+          x-field-uid: 1
+```
+* Removing a property from an object
```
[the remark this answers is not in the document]
winstonliu1111: @alakendu we came up with that list of activities in the last call where we listed them down as use cases in a notepad?  and we were crossing them out as we talk through each scenario.  include that please.

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
     * _"Number of times `session_state` changed from `up` to `down`"_ instead of
     * _"Number of times the session went from Up to Down state"_
 
+### Field-UID
```
ankur-sheth, line 318: Seems like this whole section would be better placed in the Keyword Extension section.

MODEL-GUIDE.md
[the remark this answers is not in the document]
ankur-sheth: I'd also add that showing diff's of the model (before addition of the field and after addition of the field) is important.

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
...
+
+  Filed unique identification (UID) are necessary to defined model. Please go through these use cases
```
ankur-sheth, line 320: Filed --> Field.

ankur-sheth (changes requested):

ankur-sheth (changes requested):
The `x-include` section and the example at line 146 needs to be updated.

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
...
+components:
+  schemas:  
+    Prefix.Config:
+      properties:
+        ieee_802_1qbb:
+          type: boolean
+          x-field-uid: 1
+```
+* Removing a property from an object
+  * User never use that number
```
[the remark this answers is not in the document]
alakendu: done

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
...
+  schemas:  
+    Prefix.Config:
+      x-reserved-field-uids: [1]
+      properties:
+        full_duplex_100_mb:
+          type: integer
+          x-field-uid: 2
+```
+* Adding enum value
+  * enum should replace with `x-enum`.
```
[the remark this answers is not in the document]
alakendu: done

MODEL-GUIDE.md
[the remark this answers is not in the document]
alakendu: Thanks for those valuable input

MODEL-GUIDE.md
alakendu: done

MODEL-GUIDE.md
alakendu: done

MODEL-GUIDE.md
```diff
@@ -314,6 +315,95 @@ The build script will enforce the following keyword conventions:
...
+            x-enum:
+              a:
+                x-field-uid: 1
+              b:
+                x-field-uid: 2
+              c:
+                x-field-uid: 3
+          x-field-uid: 10
+```
+* Removing enum value
```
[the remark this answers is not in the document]
alakendu: added

MODEL-GUIDE.md
```diff
@@ -301,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+
+  * Use-Case : Deprecating a Property 
+    * Model should maintain `x-field-uid: 1` when property will deprecate 
+    * `x-status` must be added with status deprecated
+
+    <img src="images/use_case_4.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Removing a Property
+    * User must never use the numbers added in `x-reserved-field-uids` again.
+    * That `x-field-uid` number should include within existing or new `x-reserved-field-uids`
```
winstonliu1111, line 344: should -> must

MODEL-GUIDE.md
```diff
@@ -301,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+    * Try to remove `ieee_802_1qbb` which have `x-field-uid: 1`. That number must be included within `x-reserved-field-uids: [1]`
+
+    <img src="images/use_case_5.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Removing an enum value
+    * User must never use the numbers added in `x-reserved-field-uids` again.
+    * That `x-field-uid` number should include within existing or new `x-reserved-field-uids`
```
winstonliu1111, line 351: should -> must

MODEL-GUIDE.md
```diff
@@ -301,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+    
+    <img src="images/use_case_3.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Deprecating a Property 
+    * Model should maintain `x-field-uid: 1` when property will deprecate 
+    * `x-status` must be added with status deprecated
+
+    <img src="images/use_case_4.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Removing a Property
```
winstonliu1111, line 342: in general on Removes, question remains, do we want things to go from 

property active -> deprecated -> removed?  or allow
property active -> removed?

the distinction here is I suspect if we go directly from property active-> removed, it implies the potential of an SDK breakeage, though serialization compatibility is maintained.  

we should clearly specify our expected lifecycle here.

winstonliu1111 (changes requested):
please see my note on 

property active -> deprecated -> removed vs. property active -> removed

MODEL-GUIDE.md
ankur-sheth: Agree.  My vote would be that we go from active --> deprecated --> removed ONLY and that we clearly document it in the guide.

MODEL-GUIDE.md
```diff
@@ -302,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+    * Try to remove `ieee_802_1qbb` which have `x-field-uid: 1`. That number must be included within `x-reserved-field-uids: [1]`
+
+    <img src="images/use_case_5.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Removing an enum value
+    * User must never use the numbers added in `x-reserved-field-uids` again.
+    * That `x-field-uid` number should include within existing or new `x-reserved-field-uids`
+    * Removing property `b` which have `x-field-uid: 2`. That number must be included within `x-reserved-field-uids: [2]`
+  
+    <img src="images/use_case_6.PNG" alt="drawing" width="500"/>
```
ankur-sheth, line 354: Can we update this example so that there is already one x-reserved-field-uid on the left side and after removing the enum we have two x-reserved-field-uid on the right side?  
That way, we provide an example of having a list of x-reserved-field-uid.

MODEL-GUIDE.md
```diff
@@ -302,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+    * That reserved field UID (e.g. 2) used by other property which was removed.
+
+    <img src="images/use_case_8.PNG" alt="drawing" width="500"/>
+
+  * x-reserved-field-uids
+
+    * When a property/response is removed, the `x-field-uid` assigned to that property/response is included in the list of `x-reserved-field-uids` so that those values are never ever re-used.
+    * Use-Case : Removing a Property
+      * User must never use the numbers added in `x-reserved-field-uids` again.
+      * That `x-field-uid` number should include within existing or new `x-reserved-field-uids`
```
ankur-sheth, line 374: I didn't understand what we're trying to say here.

MODEL-GUIDE.md
```diff
@@ -302,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+      * Try to remove `ieee_802_1qbb` which have `x-field-uid: 1`. That number must be included within `x-reserved-field-uids: [1]`
```
ankur-sheth, line 375: Try to remove --> To remove
which have --> which has

MODEL-GUIDE.md
```diff
@@ -302,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+  * Use-Case : Deprecating a Property 
+    * Model should maintain `x-field-uid: 1` when property will deprecate 
+    * `x-status` must be added with status deprecated
+
+    <img src="images/use_case_4.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Removing a Property
+    * User must never use the numbers added in `x-reserved-field-uids` again.
+    * That `x-field-uid` number should include within existing or new `x-reserved-field-uids`
+    * Try to remove `ieee_802_1qbb` which have `x-field-uid: 1`. That number must be included within `x-reserved-field-uids: [1]`
```
ankur-sheth, line 345: Try to remove --> To remove
which have --> which has

MODEL-GUIDE.md
alakendu: remove this section

MODEL-GUIDE.md
```diff
@@ -302,6 +303,79 @@ The build script will enforce the following keyword conventions:
...
+
+    <img src="images/use_case_2.PNG" alt="drawing" width="500"/>
+
+  * Use-Case : Including a Pre-Existing Object
+    * `x-include` to include only the field of the property (`..#/components/schemas/GlobalObject/properties/name`)
+    * Same property name (`uniquename`) must be used for both base and derived property. 
+    * The bundler will correctly merge the `x-include` (e.g. `type` and `pattern`) and drop the `x-include` keyword from the merged object.
+    * At the time of merge it will not include base `x-field-uid` `x-field-uid: 2` should present in the derived property.
+    
+    <img src="images/use_case_3.PNG" alt="drawing" width="500"/>
```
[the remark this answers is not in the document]
alakendu: I put this an example how it will looks like after x-include. Anyway, I have put more properties for comparison.

PR #261, opened 2022-08-18
Support for manual gateway MAC

### LINKED ISSUE #260: Need support to add Manual Gateway mac in IPv4 stack
We need a feature that is to disable resolve gateway checkbox in IPv4 stack and enter manual gateway Mac  in IPv4 stack which is the requirement for a use case in DASH project

If this can't be done, need a way to edit the mac address in the packet from the endpoint settings in traffic item

ANISH-GOTTAPU:
redocly  : https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/manual_gateway_mac/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config

resolves #260 

There is a requirement to set manual gateway MAC as per #260 

Snappi snippet
```
import snappi


def test_manual_gateway():
    api = snappi.api()
    config = api.config()

    d1, d2 = config.devices.device(name='d1').device(name='d2')

    # connected to physical port
    e1 = d1.ethernets.ethernet()[-1]
    e1.port_name = p1.name
    e1.name = 'e1'
    e1.mac = '00:01:00:00:00:01'

    # connected to lag
    e2 = d2.ethernets.ethernet()[-1]
    e2.port_name = p2.name
    e2.name = 'e2'
    e2.mac = '00:01:00:00:00:02'

    ip1, ip2 = e1.ipv4_addresses.add(), e2.ipv4_addresses.add()
    ip1.name, ip2.name = "ip1", "ip2"
    ip1.address = "10.1.1.1"
    ip1.gateway = "10.1.1.2"
    ip1.gateway_mac.value = "aa:aa:aa:aa:aa:aa"

    ip2.address = "10.1.1.2"
    ip2.gateway = "10.1.1.1"

    print(config)
```


device/ipv4.yaml
```diff
@@ -57,7 +59,23 @@ components:
...
           format: ipv4
           default: 0.0.0.0
-        
+
+    Device.Ipv4GatewayMAC:
+      description: >- 
+        Object to set Manual Gateway MAC
+      type: object
+      properties:
+        auto_resolve:
```
ajbalogh, line 68: The property name enable would be more clear than auto_resolve which would then shift the description to the address property. The address property description should reflect that the property value will be used as the resolved gateway MAC.

ANISH-GOTTAPU: I'm not sure `enable` would make more sense as gateway mac would be available once arp is done . If we add `manual_gateway_mac` and then enable would make more sense as shown below. Let me know your thoughts

```
    ip2.address = "10.1.1.2"
    ip2.gateway = "10.1.1.1"
    ip2.manual_gateway_mac.enable = True
    ip2.manual_gateway_mac.address = "bb:bb:bb:bb:bb:bb"
```

ajbalogh: @ANISH-GOTTAPU I like this second proposal.

device/ipv4.yaml
```diff
@@ -34,6 +34,8 @@ components:
             The IPv4 address of the gateway
           type: string
           format: ipv4
+        manual_gateway_mac:
```
apratimmukherjee, line 37: Does configured_gateway_mac sound better ?

ajbalogh (approved):

device/ipv4.yaml
```diff
@@ -57,7 +59,23 @@ components:
           type: string
           format: ipv4
           default: 0.0.0.0
-        
+
+    Device.Ipv4ManualGatewayMAC:
+      description: >-
```
apratimmukherjee, line 64: Instead of adding a new object , would it make sense to directly add the fields to the IPv4/IPv6 objects ( I think we just have IP/Mask/Gateway and these are 2 more ) . Something like : 
ip1.enable_configured_gw_mac  = true
 ip1.configured_gw_mac  = "aa:aa:aa:aa:aa:aa"
 vs current
ip1.manual_gateway_mac.enable = true
ip1.manual_gateway_mac.address = "aa:aa:aa:aa:aa:aa"
@ajbalogh also for comments. Not sure myself . Just a thought.

device/ipv4.yaml
```diff
@@ -57,7 +59,23 @@ components:
...
+
+    Device.Ipv4ManualGatewayMAC:
+      description: >- 
+        Object to set Manual Gateway MAC
+      type: object
+      properties:
+        enable:
+          description: >-
+            By default Manual Gateway MAC is not enabled and the resolved Gateway is used.
+            To overwrite the resolved Gateway with Manual Gateway(address), set enable to True.
```
apratimmukherjee, line 71: Maybe something like :
By default the destination MAC used for control and data packets sent from this IPv4 endpoint is learnt by sending ARP Requests ( or ND for IPv6 ) to the configured gateway IP . 
Enabling this would mean that ARP ( or ND for IPv6 resolution ) will not be used for learning MAC of  connected device.  This is needed if ARP ( or ND for IPv6) resolution is not possible in some scenario with the device where the gateway is configured.

device/ipv4.yaml
```diff
@@ -57,7 +59,23 @@ components:
...
+          type: boolean
+          default: false
+        address:
+          description: >-
+            Manual Gateway MAC.
```
apratimmukherjee, line 76: "When enable_configured_gw_mac is set to true, this user-configured MAC address will be used  for auto-filling the destination MAC address in the control and data packets sent from this IPv4 endpoint  whenever applicable. This MAC must match the MAC of the interface on the connected device where the gateway_ip is configured. " -> Something like this might be more accurate maybe ?

apratimmukherjee: Note : I am not repeating the similar comments for IPv6. Changes if done should be done symmetrically in both places.

device/ipv4.yaml
ajbalogh: I like it as an optional object as it encapsulates multiple properties, comments and functionality.

device/ipv4.yaml
ajbalogh: All of this description should be added to the container object description.

device/ipv4.yaml
apratimmukherjee: Ok , lets keep embedded object then.

device/ipv4.yaml
ANISH-GOTTAPU: updated

device/ipv4.yaml
ANISH-GOTTAPU: updated

device/ipv4.yaml
ANISH-GOTTAPU: `manual_gateway_mac` is also fine I think. @ajbalogh @ankur-sheth @ashutshkumr Can you please comment which one would be more appropriate to go with `manual_gateway_mac` or `configured_gateway_mac` ?

ashutshkumr: Here's what I suggest.

```python
 ip1.gateway_mac.learned = False
 ip1.gateway_mac.address = "aa:aa:aa:aa:aa:aa"
```

ashutshkumr: actually a better approach would be `auto` vs `value` choice.

```python
# manual config
ip1.gateway_mac.value = "aa:aa:aa:aa:aa:aa"

# auto config
nothing required
```

ashutshkumr (approved):

apratimmukherjee (commented):
Looks good to me

PR #280, opened 2022-11-02
bgp local pref changes

SouravSinhaRoy:
Model path https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/bgp_localpreference/artifacts/openapi.yaml&nocors


device/bgp/routepathattributes.yaml
```diff
@@ -22,3 +22,11 @@ components:
               x-field-uid: 1
             egp:
               x-field-uid: 2
+            incomplete:
+              x-field-uid: 3
+        local_preference:
+          description: >-
+            Local Preference attribute, for iBGP peers to select the path for the traffic leaving the AS.
```
apratimmukherjee, line 29: a) ... when multiple paths are available to the destination .
b) We can specify if larger value means higher preference or lower preference. 
c) I think we need an include_local_preference to allow user to disable inclusion of the attribute ( default value should be true ) esp. for eBGP use-case.

SouravSinhaRoy: updated,added include-local-pref and updated the descriptions

device/bgp/routepathattributes.yaml
```diff
@@ -22,3 +22,19 @@ components:
...
+        include_local_preference:
+          description: >-
+            BGP local preference attribute sent to internal peers to indicate the degree of preference 
+            for externally learned routes. It should be enabled only for internal peers.
+          type: boolean
+          default: true
+          x-field-uid: 3
+        local_preference:
+          description: >-
+            Local Preference attribute, it is used for iBGP peers to select the path for the traffic leaving the AS.
```
apratimmukherjee, line 36: Value to be set in the Local Preference attribute if include_local_preference is set to true. It is used for iBGP peers ...

result/bgpprefix.yaml
```diff
@@ -110,6 +110,8 @@ components:
               x-field-uid: 1
             egp:
               x-field-uid: 2
+            incomplete:
+              x-field-uid: 3
```
apratimmukherjee, line 114: Is this present in LI ?

apratimmukherjee (approved):
Looks good to me now.

device/bgp/routepathattributes.yaml
```diff
@@ -5,15 +5,28 @@ components:
         Configuration for advanced BGP route range settings.
       type: object
       properties:
+        include_multi_exit_discriminator:
+          description: >-
+            BGP Multi Exit Discriminator attribute sent to the peer to help in the route selection process. 
+            If set to true, the Multi Exit Discriminator attribute will be included in the route advertisement.
+          type: boolean
+          default: true
+          x-field-uid: 3
```
SuryyaKrJana, line 14: what is x-field-uid?

SuryyaKrJana (changes requested):

device/bgp/routepathattributes.yaml
ankur-sheth: @SuryyaKrJana please touch base w/ Ashutosh and others to understand what this field is and why it is required.  It is very important to get this field right in the models.

SouravSinhaRoy: x-field-uid pins a attribute to a particular index otherwise the attribute value gets overwritten with other attribute value

device/bgp/routepathattributes.yaml
```diff
@@ -5,15 +5,28 @@ components:
         Configuration for advanced BGP route range settings.
       type: object
       properties:
+        include_multi_exit_discriminator:
```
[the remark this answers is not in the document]
SouravSinhaRoy: it gives as a tester the option to include/exclude med,some bgp routes might carry med others won't

SuryyaKrJana (approved):

Issue #230, opened 2022-04-14
Support enumeration and filtering of flows


ajbalogh:
# Flow Metrics Enumeration and Filtering

A model for flow metrics enumeration and filtering which allows a user to
break enumerate a flow by identifying items of interest such as fields in the
flow as it enters a system under test (ingress), fields in the flow as it exits the
system under test (egress), transmit ports and transmit/receive port pairs.  
The model should be flexible enough to allow for future expansion.

- flow metric group configuration
- flow metric group request
- flow metric group response

## Test

1. Send a single flow of eth+vlan+ipv4 and receive eth+ipv4.
2. Expect dut to strip vlan and inject the correct ethernet type.
3. Get flow metrics enumerated by ingress ethernet source ip address and egress ethernet type.

## Configuration

1. Add unique name to packet ethernet src field.
2. Add unique name to egress_packet ethernet type field.
3. Add ingress item to flow metric groups.
4. Add egress item to flow metric groups.

## Metrics

1. Request specific flow, metric groups and metrics.
2. Response for the specific request should include metrics for the flow and any requested metric groups that have been configured.

## set_config (request) APPROACH 1 (metric_group name)

```yaml
flow:
  name: f1
  packet: # describe the shape of the packet as it exits the otg system enters the system under test
    ethernet:
      src:
        metric_group: f1-eth-src # provide a unique name for the field so that it can be referenced in flow.metrics
    vlan:
    ipv4:
  egress_packet: # describe the shape of the packet as it exits the system under test and enters the otg system
    ethernet:
      ether_type:
        metric_group: f1-eth-type # provide a name for the field so that it can be referenced in flow.metrics
  metrics:
    enable: true
    metric_groups: # configuration container for all metric groups
      ingresses:
        - field_name: f1-eth-src
      egresses:
        - field_name: f1-eth-type
          field_offset: 0
          field_length: 16
```

## set_config (request) APPROACH 2 (nested metric_group)

```yaml
flow:
  name: f1
  packet: # describe the shape of the packet as it exits the otg system enters the system under test
    ethernet:
      src:
        metric_group: 
          name: f1-eth-src 
    vlan:
    ipv4:
  egress_packet: # describe the shape of the packet as it exits the system under test and enters the otg system
    ethernet:
      ether_type:
        metric_group: 
          name: f1-eth-type 
          field_offset: 0
          field_length: 16
  metrics:
    enable: true
```

## get_metrics (request)

```yaml
flow:
[...]

PR #281, opened 2022-11-07
Flow metrics enumeration

ajbalogh:
Addresses issue [Support flow egress metrics enumeration](#230)

## Proposed Changes
  - proposal to standardize enumeration that will apply to both ingress and egress packet headers which allows for one version of packet header and field models
  - this will require a new version of openapiart where the `x-field-pattern metric_groups` feature generates the following code per field (metric_group will be supported but eventually deprecated):
  ```yml
  # source model
  Flow.Ipv4:
    properties:
      src:
        x-field-pattern:
          features: [metric_groups]
  
  # openapiart generated model
  Pattern.Flow.Ipv4.Src:
    properties:
      metric_groups:
        type: array
        items:
          type: object
          required: [name]
          properties:
            name: 
              type: string
            offset: 
              type: integer
              default: 0
              min: 0
              max: "field length - 1"
            length: 
              type: integer
              default: "field length"
              min: 1
              max: 32
  ``` 
  - [Proposed model changes here](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/egress_metric_group/artifacts/openapi.yaml&nocors)

## Sample Code
```python
import snappi

api = snappi.api(
    location="127.0.0.1:5001",
    transport=snappi.Transport.GRPC,
    verify=False,
)

config = api.config()
p1, p2 = config.ports.port(name="p1").port(name="p2")

# create a flow
flow = config.flow.add(name="f1")
flow.tx_rx.port.tx_name = p1.name
flow.tx_rx.port.rx_name = p2.name
flow.duration.fixed_packets.packets = 10000
flow.rate.pps = 1000

# the ingress packet to the system under test will include the following
# protocol template headers
i_eth, i_vlan, i_ipv4 = flow.packet.ethernet().vlan().ipv4()
i_vlan.priority.value = 3
i_ipv4.src.metric_groups.add(name="ingress-ipv4-src", offset=0, width=32)
i_ipv4.src.values = ["1.1.1.1", "1.1.1.2", "1.1.1.3"]
i_ipv4.dst.value = "1.1.2.1"

# the egress packet from the system under test will include the
# following protocol template headers with the expectation that
# the system under test will strip off the vlan header
_, e_ipv4 = flow.metrics.egress.ethernet().ipv4()

# the egress packet can be configured to enumerate the ipv4 version field
# this should be at offset 112 for 4 bits
# the egress.ethernet().ipv4().version field makes it easy to configure the
# field without having to calculate the offset from the start of packet
# the enumeration object metric_groups is an array to allow for multiple
# enumerations in a single field
e_ipv4.version.metric_groups.add(name="egress-ipv4-version", offset=0, length=4)

# retrieve all flow metrics enumerated by egress and ingress metric groups
request = api.metrics_request()
request.flow.flow_names = []
response = api.get_metrics(request)

# printing the metrics response should yield the following
```
If the VLAN header is successfully stripped the egress-ipv4-version metric group value will have the value 0x04 in the response
|flow|tx-port|rx-port|ingress-ipv4-src|egress-ipv4-version|octets-out|octets-in|
|-|-|-|-|-|-|-|
|f1|p1|p2|||10000|10000|
|f1|p1|p2|1.1.1.1|0x04|3334|3334|
|f1|p1|p2|1.1.1.2|0x04|3333|3333|
|f1|p1|p2|1.1.1.3|0x04|3333|3333|

If the VLAN header is NOT successfully stripped the egress-ipv4-version will have the value 0x06 from the VLAN priority in the response
|flow|tx-port|rx-port|ingress-ipv4-src|egress-ipv4-version|octets-out|octets-in|
|-|-|-|-|-|-|-|
|f1|p1|p2|||10000|10000|
|f1|p1|p2|1.1.1.1|0x06|3334|3334|
|f1|p1|p2|1.1.1.2|0x06|3333|3333|
|f1|p1|p2|1.1.1.3|0x06|3333|3333|

Alternate methods of configuring egress metric groups
```python
# alternate approach of using an offset from the start of a protocol template
# by using the custom packet header
_, e_custom = flow.metrics.egress.ethernet().custom()
e_custom.field.metric_groups.add(name="egress-ipv4-version", offset=0, length=4)

# alternate approach of using an offset from the start of the packet by using
# the custom packet header
e_custom = flow.metrics.egress.custom()
e_custom.field.metric_groups.add(name="egress-ipv4-version", offset=112, length=4)

# use the custom packet header for variable length headers
# e.g., if ipv4 has options present
_, _, e_custom = flow.metrics.egress.ethernet().ipv4().custom()
e_custom.field.metric_groups.add(name="egress-ipv4-options", offset=0, length=32)
```

Alternate methods of requesting flow metric results
```python
# retrieve only disaggregated flow metrics
# will not return the aggregate row
request = api.metrics_request()
request.flow.flow_names = []
request.flow.metric_groups = [
  {
    metric_group_name: "egress-ipv4-version",
    filters: []
  },
  {
    metric_group_name: "ingress-ipv4-src",
    filters: []
  }
]
response = api.get_metrics(request)
```


jasdeep-hundal:
The protocol header approach is definitely cleaner, but I'm wondering about:
- How does it handle variable lengths of headers? (Such as if options are set in an IPv4 header)
- Would we be able to track/validate fields in a protocol that isn't supported in OTG yet? (Eg. could we just specify offset in an ethernet packet?)

result/flow.yaml
```diff
@@ -45,43 +45,28 @@ components:
...
-            If the items is empty then no disaggregation will be done. 
+            A list of filters that can be applied to the metric group name.
+            By default all values will be included in the flow metric results.
           type: array
-          default: []
           items:
             type: string
-          x-constraint:
-          - "/components/schemas/Flow/properties/packet/../metric_group"
+            format: hex
```
winstonliu1111, line 67: hex?  

are there still x-constraints for the name property?

flow/egress/egress.yaml
```diff
@@ -0,0 +1,238 @@
+components:
+  schemas:
+    Egress.Header:
+      description: >-
+        List of all egress packet headers
```
ankur-sheth, line 5: Need to specify somewhere that this is egress from the DuT/SuT point of view.

flow/egress/egress.yaml
```diff
@@ -0,0 +1,238 @@
...
+      type: object
+      properties:
+        choice:
+          description: >-
+            The type of egress packet header
+          type: string
+          default: custom
+          x-field-uid: 1
+          x-enum:
+            custom:
```
ankur-sheth, line 15: Do we not need 'vxlan', 'gre', 'tcp', 'udp' also?

flow/egress/egress.yaml
```diff
@@ -0,0 +1,238 @@
...
+        metric_group:
+          description: >-
+            A unique name is used to indicate to the system that the field may
+            extend the metric row key and create an aggregate metric row for
+            every unique value. To have metric group columns appear in the flow
+            metric rows the flow metric request allows for the metric_group
+            value to be specified as part of the request.
+          type: string
+          x-field-uid: 1
+        offset:
```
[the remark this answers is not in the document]
ankur-sheth: Is the offset in bits or bytes?  I'm assuming the former, let's clarify in description.

flow/metrics.yaml
```diff
@@ -30,6 +30,17 @@ components:
...
           x-field-uid: 4
+        egress:
+          description: |-
+            Enumerate flow metrics by describing the shape of the packet as it
+            exits the system under test and assigning unique names to
+            individual packet header fields.
+          type: array
+          items:
+            $ref: './egress/egress.yaml#/components/schemas/Egress.Header'
+          x-field-uid: 8
```
ankur-sheth, line 41: What's the reason for value 8 and not 5?   Were the values 5, 6 & 7 used earlier?

result/flow.yaml
```diff
@@ -45,43 +45,28 @@ components:
               frames_rx_rate:
                 x-field-uid: 7
           x-field-uid: 3
-    Flow.MetricGroup.Request:
+
+    Flow.MetricGroup.Filter:
```
ankur-sheth, line 49: Are these chances breaking backwards compatibility?

flow/metrics.yaml
```diff
@@ -30,6 +30,17 @@ components:
             Latency metrics.
           $ref: '#/components/schemas/Flow.Latency.Metrics'
           x-field-uid: 4
+        egress:
```
apratimmukherjee, line 33: Another comment related to how user sees results: Say we have a flow 10 sub-flows , 8 have MPLS header and 2 dont have MPLS header and user does egress tracking by MPLS 3 bit TOS/EXP bits . It looks like as per the example, even the 2 sub-flows which dont have MPLS header at all will get added to some bucket arbitrarily based on values in the 3 bits where MPLS EXP would have been if it had been present. Isn't this behaviour definition problematic  ? Should it not be that if packet does not confirm to packet shape as defined ( excluding custom ) , only those packets should be filtered from aggregate flow which matches the defined egress packet shape ?

ashutshkumr:
I have a concern regarding defining separate set of protocol headers for egress tracking. Whenever there's a modification in packet headers, one would have to be extremely cautious of making the same change in egress tracking.

So TL;DR, I was hoping to explore another possibility where users wouldn't have to work with two different packet header descriptions (flow.packet vs flow.metrics.egress).

One such approach could be following, assuming that most use cases of egress tracking shall fall under category of header-stripping (vlan and tunneling protocols) and field value replacement (e.g. QOS).

NOTE: This is just a brain-dump and I haven't really evaluated all possible scenarios while providing the snippet

```python
# USECASE 1: packet for which TOS precedence value will change
eth, ip, tcp = flow.packet.ethernet().ipv4().tcp()
# set IP src/dst
ip.priority.tos.precedence.values = [1, 4, 7]
# configure ingress tracking for IP priority
ip.priority.metric_group.ingress.name = "ingress-ipv4-priority"
# configure egress tracking for IP priority
ip.priority.metric_group.egress.name = "egress-ipv4-priority"

# USECASE 2: packet for which VLAN header shall be stripped
eth, vl, ip = flow.packet.ethernet().vlan().ipv4()
# annotate VLAN header as stripped
vl.priority.value = 3
vl.egress_trackable = False
# set IP src/dst
ip.dst.value = "1.1.2.1"
ip.src.values = ["1.1.1.1", "1.1.1.2", "1.1.1.3"]
# configure ingress tracking for IP src
ip.src.metric_group.ingress.name = "ingress-ipv4-src"
# configure egress tracking for IP version (without needing to provide offset / length)
ip.src.metric_group.egress.name = "egress-ipv4-version"

# optionally, to provide custom offset / length
ip.src.metric_group.egress.custom.bit_offset= 0
ip.src.metric_group.egress.custom.bit_length = 4
# if False, offset shall be calculated compared to start of frame
ip.src.metric_group.egress.custom.relative_offset = True
```

ashutshkumr:
> While inlining the egress tracking options in the ingress packet reduces model overhead it doesn't provide the user with a model of the egress packet. The intent of the .metrics.egress... is to do just that.

If I'm assuming correctly regarding the stripped header use cases, the egress packet will mostly be a subset (with some field modifications) of the original packet headers right? Then what's wrong with annotating that a certain header won't be there in the egress like so, with `egress_trackable=False`?

```python
# USECASE 2: packet for which VLAN header shall be stripped
eth, vl, ip = flow.packet.ethernet().vlan().ipv4()
# annotate VLAN header as stripped
vl.priority.value = 3
vl.egress_trackable = False
```

ashutshkumr:
Thanks @ajbalogh and @apratimmukherjee, it's now clearer that egress packets won't be limited to only either insertion or stripping; instead it could consist of combination of both stripping and insertion (e.g. get rid of vlan header and insert mpls header), which invalidates my initial proposal.

But in hopes to still explore other options of reusability, here's another shot :sweat_smile:.

Would it make sense to introduce a new choice for header fields (alongside `[value, values, increment, decrement]`), called `egress_metric` ? We could call it something shorter.

Reason to do that would be two-fold:
- We could reuse same `type` for packet headers in `flow.packet` and `flow.metrics.egress_packet` (rephrasing `flow.metrics.egress` for clarity)
- We would still retain mutual exclusion between `configuring egress metric` vs `configuring field values`.

The example now looks more similar to author's initial proposal (renamed `metric_group` to `ingress_metric_name` for consistency).

```python
# USECASE 1: packet for which TOS precedence value will change
eth, ip, tcp = flow.packet.ethernet().ipv4().tcp()
# set IP src/dst
ip.priority.tos.precedence.values = [1, 4, 7]
# configure ingress metric for IP priority
ip.priority.ingress_metric_name = "ingress-ipv4-priority"

# configure egress metric for IP priority (without needing to provide offset / length)
_, eip, _tcp = flow.metrics.egress_packet.ethernet().ipv4().tcp()
eip.priority.egress_metric.name = "egress-ipv4-priority"


# USECASE 2: packet for which VLAN header shall be stripped
eth, vl, ip = flow.packet.ethernet().vlan().ipv4()
vl.priority.value = 3
# set IP src/dst
ip.dst.value = "1.1.2.1"
ip.src.values = ["1.1.1.1", "1.1.1.2", "1.1.1.3"]
# configure ingress metric for IP src
ip.src.ingress_metric_name = "ingress-ipv4-src"

# configure egress metric for IP version (without needing to provide offset / length)
_, eip = flow.metrics.egress_packet.ethernet().ipv4()
eip.src.egress_metric.name = "egress-ipv4-version"

# optionally, to provide custom offset / length
eip.src.egress_metric.offset.bit_start = 0
eip.src.egress_metric.offset.bit_length = 4

# if True, offset shall be calculated from start of frame (default = True)
eip.src.egress_metric.offset.from_start_of_frame = True
```

anjan-keysight:
The offset and length per egress packet header fields, assumes the tracking bits are contiguous, but technically we can support non-contiguous tracking bits. And since, currently, there is a limit on the number of tracking bits we can set, would it make more sense to have an array of offset and length per header, or a bitmask.

flow/metrics.yaml
```diff
@@ -8,7 +8,7 @@ components:
         enable:
```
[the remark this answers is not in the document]
PrasenjitAdhikary: Before finalizing the OTG API model, can also finalize the gNMI model as well, to avoid any conflict between these two models and the possibility of rework in the future?

ajbalogh:
superseded by #312

PR #284, opened 2022-11-16
Transmit add in port metrics response

shramroy:


ashutshkumr (changes requested):
Please do not alter x-field-uid values of other fields.

shramroy:
> Please do not alter x-field-uid values of other fields.

@ashutshkumr I have modified it. Kindly check

ashutshkumr (approved):

PR #290, opened 2022-12-07
Rename DeviceOptions to ProtocolOptions

ashutshkumr:
This PR renames the newly introduced attribute in https://github.com/open-traffic-generator/models/pull/286. See the documentation [here](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-autostart/artifacts/openapi.yaml&nocors).

The reason being, there are other protocols like LLDP / LAG-LACP which are defined outside the `device` hierarchy. Hence renaming `DeviceOptions` to `ProtocolOptions` is more inclusive.

A new global option for devices has been introduced in `config` to indicate whether all underlying resources for configured devices shall be created and corresponding protocol session negotiation shall be negotiated upon `set_config` API call. Corresponding gosnappi snippet would look like so:

   ```go
   // do not initiate protocol session negotiation upon set_config API call
   config.Options().ProtocolOptions().SetAutoStartAll(false)
   ```

Default is `true`.


device/device.yaml
```diff
@@ -68,18 +68,18 @@ components:
...
       type: object
       properties:
         auto_start_all:
           description: >-
-            When set to true, all underlying resources for configured devices 
-            shall be created and corresponding protocol session negotiation shall 
-            be initiated. Otherwise, when set to false, corresponding protocol 
-            session negotiation will need to be initiated using a separate 
-            set_protocol_state API call.
+            When set to true, all underlying resources for configured protocols
```
apratimmukherjee, line 78: I understand that when L47 state-full flows become part of OTG, then user will specifically need to do something similar to StartTransmit explicitly and this will not in ambit of auto_start_all=true option.

apratimmukherjee (approved):
LGTM

PR #304, opened 2023-02-22
BGP Graceful Restart draft version

SuryyaKrJana:


device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,60 @@
...
+          description: |-
+            This time (in seconds) specifies how long stale information (for the AFI/SAFI) 
+            may be retained (possibly in conjunction with the period specified by the "Restart Time" 
+            in the Graceful Restart Capability, if present).
+          type: integer
+          format: uint32
+          maximum: 16777216
+          default: 10
+          x-field-uid: 4
+        act_as_restarted:
```
apratimmukherjee, line 50: This should probably be removed. This is very IxNetwork specific and not needed with trigger being provided.

control/control.yaml
```diff
@@ -213,3 +213,28 @@ components:
...
+            x-constraint will be affected.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Device.Bgp/properties/name'
+          x-field-uid: 1
+        restart_time:
+          description: >-
+            This is the estimated time (in seconds) it will take for the
```
apratimmukherjee, line 235: This is not what is going in the BGP PDU but rather the time after which selected the BGP peer(s) will restart the connection with Restart Bit set in GR Capability to emulate Graceful Restart. The max value might be better to have as something like 1 hr for LLGR use-cases for example.

result/bgpv6.yaml
```diff
@@ -170,4 +174,16 @@ components:
             openconfirm:
               x-field-uid: 5
             established:
-              x-field-uid: 6
\ No newline at end of file
+              x-field-uid: 6
+        gr_restart_attempted:
+          description: |-
+            Number of Graceful Restarts attempted by BGP Peer per session.
```
apratimmukherjee, line 180: For these , good to specify it is for peer graceful restarts probably when we are acting as Helper. ( assuming that is the case ) . Also good to specify exactly when they will be incremented. It is not very obvious to me.

apratimmukherjee (commented):
Quick initial first-round comments.

control/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,33 @@
...
+    Bgp.GraceFulRestart:
+      description: >-
+        Initiates BGP Graceful Restart process for the selected BGP peers.
+      type: object
+      properties:
+        peer_names:
+          description: >-
+            The names of device BGP peer objects to control.
+            If no names are specified then all BGP peer objects that match the
+            x-constraint will be affected.
```
PrasenjitAdhikary, line 18: x-constraint is applicable even when specific names are specified, and implicit for the peers in this context. So, can we remove mentioning x-constraint only for no-names?

result/bgpv6.yaml
```diff
@@ -59,6 +59,10 @@ components:
                 x-field-uid: 14
               fsm_state:
                 x-field-uid: 15
+              gr_restart_attempted:
+                x-field-uid: 16
+              gr_restart_failed:
+                x-field-uid: 17
```
PrasenjitAdhikary, line 65: gr_restart_routes_received is missing.

result/bgpv4.yaml
```diff
@@ -58,6 +58,10 @@ components:
                 x-field-uid: 14
               fsm_state:
                 x-field-uid: 15
+              gr_restart_attempted:
+                x-field-uid: 16
+              gr_restart_failed:
+                x-field-uid: 17
```
PrasenjitAdhikary, line 64: gr_restart_routes_received is missing from the list.

result/bgpv4.yaml
```diff
@@ -169,4 +173,24 @@ components:
...
+        gr_restart_attempted:
+          description: |-
+            Number of Graceful Restarts attempted by the neighbor of the BGP peer. This is incremented
+            when a Open Message received indicating that neighbor has restarted.
+          type: integer
+          format: uint32
+          x-field-uid: 17
+        gr_restart_failed:
+          description: |-
+            Number of Graceful Restarts that were attempted by BGP Peer, however the BGP session
```
apratimmukherjee, line 186: _The number of times that the BGP session could not be re-established within the Graceful Restart time after Graceful Restart was attempted by the neighbor of the BGP peer._ might be better. 
However, confused what happens if OPEN itself is not recieved , attempted stat will not be incremented but this will be ? So we can get _gr_restart_failed_  > _gr_restart_attempted_ ?

result/bgpv4.yaml
```diff
@@ -169,4 +173,24 @@ components:
...
+            was not re-established within the Graceful Restart Time.
+          type: integer
+          format: uint32
+          x-field-uid: 18
+        gr_restart_routes_received:
```
apratimmukherjee, line 191: I think we wanted to remove this for now till we have more clarity on when this is useful as a metric.

result/bgpv6.yaml
```diff
@@ -170,4 +174,24 @@ components:
...
+          format: uint32
+          x-field-uid: 17
+        gr_restart_failed:
+          description: |-
+            Number of Graceful Restarts that were attempted by the BGP Peer, however the BGP session
+            was not re-established within the Graceful Restart Time.
+          type: integer
+          format: uint32
+          x-field-uid: 18
+        gr_restart_routes_received:
```
apratimmukherjee, line 192: Same as for bgpv4

SuryyaKrJana: ok

result/bgpv4.yaml
```diff
@@ -169,4 +173,24 @@ components:
...
+          description: |-
+            Number of Graceful Restarts attempted by the neighbor of the BGP peer. This is incremented
+            when a Open Message received indicating that neighbor has restarted.
+          type: integer
+          format: uint32
+          x-field-uid: 17
+        gr_restart_failed:
+          description: |-
+            Number of Graceful Restarts that were attempted by BGP Peer, however the BGP session
+            was not re-established within the Graceful Restart Time.
```
SuryyaKrJana, line 187: ok

result/bgpv4.yaml
SuryyaKrJana: I dont have the answer, pls let me know want to keep or not

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,51 @@
...
+            This capability MUST be advertised in conjunction with the Graceful Restart 
+            capability.
+          type: boolean
+          default: false
+          x-field-uid: 3
+        stale_time:
+          description: |-
+            This time (in seconds) specifies how long stale information (for the AFI/SAFI) 
+            may be retained (possibly in conjunction with the period specified by the "Restart Time" 
+            in the Graceful Restart Capability, if present). This 3 Bytes field and
```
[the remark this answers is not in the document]
SuryyaKrJana: chk now

result/bgpv4.yaml
apratimmukherjee: Let us keep for now. I guess we need to 
a) check if this theoretical scenario is actually true or not with actual emulated test. 
b)  check for current OC bgp metrics , whether it provides some counters for GR restarts for helper node [ unlikely there will be anything for restarting node ] , and 
c) work out if what have today makes sense or needs some change of behaviour in the imp. for these metric s to make it useful for end-user for validation purposes.  -->  I think having it seems to be better than not having it , even in present form.

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,50 @@
...
+          type: integer
+          format: uint16
+          maximum: 4096
+          default: 45
+          x-field-uid: 2
+        enable_llgr:
+          description: |-
+            If enabled, the "Long-lived Graceful Restart Capability", or "LLGR Capability"
+            will be advertised.
+            This capability MUST be advertised in conjunction with the Graceful Restart
```
SouravSinhaRoy, line 35: If must do e have to expose it ?

apratimmukherjee:
Need to add the trigger API in the example.

device/bgp/bgpgracefulrestart.yaml
apratimmukherjee: The ATE test uses it , so I think we need to also expose it.

control/bgp.yaml
```diff
@@ -87,3 +92,27 @@ components:
...
+            The names of device BGP peer objects to control.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Device.Bgp/properties/name'
+          x-field-uid: 1
+        restart_time:
+          description: >-
+            This is the time (in seconds) after which the selected BGP peers will initiate the
```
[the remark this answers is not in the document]
apratimmukherjee: it should be peers I think ..not peer.

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,50 @@
...
+        The Graceful Restart Capability (RFC 4724) is a BGP capability that can be
+        used by a BGP speaker to indicate its ability to preserve its forwarding
+        state during BGP restart.
+        The Graceful Restart (GR) capability is advertised in OPEN messages sent between BGP peers.
+        After a BGP session has been established, and the initial routing update has been completed, 
+        an End-of-RIB (Routing Information Base) marker is sent in an UPDATE message to convey information 
+        about routing convergence.
+      type: object
+      properties:
+        enable_gr:
```
SouravSinhaRoy, line 14: If GracefulRestart object is included then I think it means graceful restart is enabled then shall we require "enable_gr"

control/bgp.yaml
```diff
@@ -87,3 +92,27 @@ components:
...
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Device.Bgp/properties/name'
+          x-field-uid: 1
+        restart_time:
+          description: >-
+            This is the time (in seconds) after which the selected BGP peers will initiate the 
+            Graceful restart by sending the Open Message with Restart bit set.
```
apratimmukherjee, line 113: with Restart State bit set in the Graceful Restart capability .

apratimmukherjee (approved):
Looks good to me. One suggestion for documentation improvement mentioned + need to add API trigger in example before merging,

device/bgp/bgpgracefulrestart.yaml
SouravSinhaRoy: ok

device/bgp/bgpgracefulrestart.yaml
SuryyaKrJana: Yes, its true; using the has property we could avoid the enable_gr boolean flag.
Then a user needs to configure then it in the following way
bgpPeerObj.SetGracefulRestart() in contrast to bgpPeerObj.SetGracefulRestart.SetEnableGR(true).
We find the second option is more suitable option from script readable perspective. 
Another reason that's backed the second option, in future, we want to disable Graceful Restart option on the fly.

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,50 @@
...
+          default: false
+          x-field-uid: 1
+        restart_time:
+          description: >-
+            This is the estimated time (in seconds) it will take for the
+            BGP session to be re-established after a restart.  This can be
+            used to speed up routing convergence by its peer in case
+            the BGP speaker does not come back after a restart.
+          type: integer
+          format: uint16
```
apratimmukherjee, line 27: As per Ashutosh , we cannot have uint16 so we should use uint32 instead  probably. Need to check all fields where fmt has been used. Also, this model should be merged with main only after otgen supports new fmts uint32 and uint64 used in this PR

apratimmukherjee (changes requested):

device/bgp/bgpgracefulrestart.yaml
SuryyaKrJana: done

apratimmukherjee (approved):

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -0,0 +1,50 @@
...
+          default: false
+          x-field-uid: 3
+        stale_time:
+          description: |-
+            This time (in seconds) specifies how long stale information (for the AFI/SAFI) 
+            may be retained. This is a three byte field and is applicable 
+            only if 'enable_llgr' is set to 'true'.
+          type: integer
+          format: uint32
+          maximum: 16777216
```
PrasenjitAdhikary, line 47: The range should be 0-16777215. Max value looks to be corrected.

PrasenjitAdhikary (approved):
LGTM

device/bgp/bgpgracefulrestart.yaml
SuryyaKrJana: yes

device/bgp/bgpgracefulrestart.yaml
```diff
@@ -39,7 +39,7 @@ components:
           x-field-uid: 3
         stale_time:
           description: |-
-            This time (in seconds) specifies how long stale information (for the AFI/SAFI) 
+            Duration (in seconds) specifies how long stale information (for the AFI/SAFI)
```
apratimmukherjee, line 42: specified -> specifying . Or Duration (in seconds) for which long stale information ....

apratimmukherjee (approved):

SuryyaKrJana:
Merging BGP gracefully Restart to Master

PR #312, opened 2023-03-16
Support enumeration and filtering of flows #230

dipendughosh:
Fixes https://github.com/open-traffic-generator/models/issues/230

View proposed changes in [documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/test_branch_egress_ingress/artifacts/openapi.yaml&nocors)

## What has changed (and why) ?
- #### How to distinguish between egress and ingress flows-
	* Previously flow.packet were used to configure shape of ingress packets only
	* There was no clear distinction between ingress and egress packet shapes, and no way to configure latter
	* Going forward the distinction between ingress and egress flows will be done with `flow.ingress_packet` and `flow.egress_packet`. 
	* `flow.packet` shall be deprecated.
- #### How to track the egress and ingress flows-
	* Under packet header configuration for each field,
		* `metric_groups` has been replaced with `metric_tags` containing a list of `metric_tag` consisting of properties name, offset and length.
	* Under flow metrics request,
		* `metric_groups` has been replaced with `metric_tags` containing a list of filter consisting of tag name and list of corresponding values associated with that tag name

### What's pending ?
- Description needs to be updated properly. Need feedback from reviewers.

### Backwards Compatibility
- No breakage in existing tests - `flow.packet` is deprecated and `metric_groups` was never implemented.
- snappi UTs needs update to replace usage of metric_groups with metric_tags and use ingress packet configuration

### Example

```go
api := gosnappi.NewApi()
config := api.NewConfig()

ptx := config.Ports().Add().SetName("ptx").SetLocation("eth1")
prx := config.Ports().Add().SetName("prx").SetLocation("eth2")

f1 := config.Flows().Add().SetName("f1")

f1.TxRx().Port().SetTxName(ptx.Name()).SetRxName(prx.Name())
f1.Metrics().SetEnable(true)
f1.Size().SetFixed(128)
f1.Rate().SetPps(100)

// Configure shape of packet to be sent to DUT interface connected to port ptx
inEthF1 := f1.IngressPacket().Add().Ethernet()
inVlanF1 := f1.IngressPacket().Add().Vlan()
inIpF1 := f1.IngressPacket().Add().Ipv4()

inEthF1.Src().SetValue("00:00:00:00:00:AA")

inVlanF1.Id().SetValues([]int32{1, 2})
// Track ingress VLAN ID
inVlanF1.Id().MetricTags().Add().SetName("inVlanF1Id").SetOffset(10).SetLength(2)

inIpF1.Src().SetValue("1.1.1.1")
inIpF1.Dst().SetValues([]string{"1.1.1.3", "1.1.1.4"})

// These values will be transformed by DUT into priorities 5 and 6 respectively
inIpF1.Priority().Tos().Precedence().SetValues([]int32{2, 3})
// Track ingress IPv4 TOS precedence
inIpF1.Priority().Tos().Precedence().MetricTags().Add().SetName("inIpF1Tos")

// Configure packet shape to be expected as it comes out of DUT interface connected
// to port prx; VLAN header will be stripped
f1.EgressPacket().Add().Ethernet()
egIpF1 := f1.EgressPacket().Add().Ipv4()

// Track egress IPv4 destination address
egIpF1.Dst().MetricTags().Add().SetName("egIpF1Dst").SetOffset(28).SetLength(2)

// Track egress IPv4 TOS precedence
egIpF1.Priority().Tos().Precedence().MetricTags().Add().SetName("egIpF1Tos")

api.SetConfig(config)

cs := api.NewControlState()
cs.Traffic().FlowTransmit().SetState(gosnappi.StateTrafficFlowTransmitState.START)
api.SetControlState(cs)

mr := api.NewMetricsRequest()
mr.Flow()
ms, _ := api.GetMetrics(mr)

for _, row := range ms.FlowMetrics().Items() {
	if len(row.MetricTags().Items()) > 0 {
		fmt.Println("Flow metrics with disaggregated counters", row)
	} else {
		fmt.Println("Flow metrics with aggregated counters", row)
	}
}
```

### Output without using any query filters

| Flow | FramesTx | FramesRx | MetricTags                                                                                                                                                                        | Comments                                                                                               |
|------|----------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| f1   | 100      | 100      | nil                                                                                                                                                                               | Aggregated counters for configured flow                                                                |
| f1   | nil      | 50       | [{"name":"inVlanF1Id","value":"0x1"},<br>{"name":"inIpF1Tos","value":"2"},<br>{"name":"egIpF1Tos","value":"5"},<br>{"name":"egIpF1Dst","value":"0x3"}] | Disaggregated counters for Nth packet<br>in the flow where N is an odd number                          |
| f1   | nil      | 49       | [{"name":"inVlanF1Id","value":"0x2"},<br>{"name":"inIpF1Tos","value":"3"},<br>{"name":"egIpF1Tos","value":"6"},<br>{"name":"egIpF1Dst","value":"0x4"}] | Disaggregated counters for Nth packet<br>in the flow where N is an even number                         |
| f1   | nil      | 1        | [{"name":"inVlanF1Id","value":"0x2"},<br>{"name":"inIpF1Tos","value":"3"},<br>{"name":"egIpF1Tos","value":"7"},<br>{"name":"egIpF1Dst","value":"0x4"}] | Disaggregated counters for packets in<br>the flow which had incorrect egress<br>TOS precedence value 7 |

### Output upon using query filters

```go
mr := api.NewMetricsRequest()
mr.Flow().MetricTags().Add().SetName("inVlanF1Id")
ms, _ := api.GetMetrics(mr)
```

| Flow | FramesTx | FramesRx | MetricTags                                                                                                                                                                        | Comments                                                                                               |
|------|----------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| f1   | 100      | 100      | nil                                                                                                                                                                               | Aggregated counters for configured flow                                                                |
| f1   | nil      | 50       | [{"name":"inVlanF1Id","value":"0x1"}}] | Disaggregated counters for Nth packet<br>in the flow where N is an odd number                          |
| f1   | nil      | 50       | [{"name":"inVlanF1Id","value":"0x2"}] | Disaggregated counters for Nth packet<br>in the flow where N is an even number                         |


result/flow.yaml
```diff
@@ -15,12 +15,14 @@ components:
           x-constraint:
           - "/components/schemas/Flow/properties/name"
           x-field-uid: 1
-        metric_groups:
+        metric_tags:
```
PrasenjitAdhikary, line 18: How to fetch only ingress / egress stats?

result/flow.yaml
```diff
@@ -173,17 +159,17 @@ components:
...
           description: |-
-            Name of packet field metric group
+            Name of packet field metric tag
           type: string
           x-field-uid: 1
         value:
           description: |-
-            Value of named packet field metric group
+            Value of named packet field metric tag
           type: string
```
PrasenjitAdhikary, line 173: In the request, tag value is in hex format. Do we expect here a different format, or it should match with request filter.

flow/flow.yaml
```diff
@@ -29,27 +29,47 @@ components:
           minItems: 1
           items:
             $ref: './packet-headers/header.yaml#/components/schemas/Flow.Header'
+          x-status:
+            status: deprecated
+            information: |- 
+              Use flow.ingress_packet instead.
           x-field-uid: 2
+        ingress_packet:
```
apratimmukherjee, line 37: From previous review:  https://github.com/open-traffic-generator/models/pull/281
We have a scenario where for Ingress tracking, it is sometimes useful if we have a generic Flow Descriptor field which contains autogenerated unique string describing endpoint and then user can track Tx/Rx/Drop per such sub-flow in one flow/stream. This is useful for example when 10000 RSVP tunnels are aggregated into single stream where tracking by MPLS label is not very useful ( assigned by DUT and difficult to track MPLS label -> ingress tunnel + MPLS label can change at run-time e.g. Dynamic MPLS ). So a auto tracking option where sub-flows are automatically assigned names and optional per sub-flow tracking is provided can be useful. We could create one traffic item per tunnel but that gets limited by max flows/stream per port ( say mostly upto 256 for h/w) . This could also be applicable where tracking is done by autogenerated names .

flow/flow.yaml
```diff
@@ -29,27 +29,47 @@ components:
...
+          description: |-
+            Ingress packet
+          type: array
+          minItems: 1
+          items:
+            $ref: './packet-headers/header.yaml#/components/schemas/Flow.Header'
+          x-field-uid: 3
+        egress_packet:
```
apratimmukherjee, line 45: Egress packet headers continue to have value/valuelist etc. which is not applicable for script. Would be good to have a way  to re-use ingress headers but leave out unused fields from the bundled yaml.

apratimmukherjee (dismissed):

PrasenjitAdhikary:
In the example, we need a few scenarios: (a) Tracking enabled for ingress fields (b) Tracking enabled for egress fields (c) Tracking enabled for both types of fields. In all 3 situations, what are the expected stat columns & non-empty value fields, so that it matches customer expectation of drill-down & aggregated stats and Tx & Rx stats.
Also in the example, for the ingress field, enable tracking for the entire field (IP or MAC type field) - which is the most common use case for ingress tracking. In such a scenario, how to specify metric requests for retrieving stats for a specific IP address, and what will be the example filtered result (including the format for both request/response)?

parthpower:
it would be easier if there was a `choice` to pick either an offset+length or reference it from predefined header field from the flow object

result/flow.yaml
```diff
@@ -45,47 +47,31 @@ components:
               frames_rx_rate:
                 x-field-uid: 7
           x-field-uid: 3
-    Flow.MetricGroup.Request:
+    Flow.MetricTag.Filter:
```
ankur-sheth, line 50: Does changing Flow.MetricGroup.Request to Flow.MetricTag.Filter  not break backwards compatibility?

result/flow.yaml
```diff
@@ -105,12 +91,12 @@ components:
...
           type: string
           x-field-uid: 3
-        metric_groups:
+        metric_tags:
           description: |-
             Flow disaggregation groups.
           type: array
           items:
-            $ref: '#/components/schemas/Flow.MetricGroup'
+            $ref: '#/components/schemas/Flow.MetricTag'
```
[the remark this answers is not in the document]
ankur-sheth: Same question about backwards compatibliity.

ankur-sheth (changes requested):

result/flow.yaml
```diff
@@ -15,12 +15,14 @@ components:
           x-constraint:
           - "/components/schemas/Flow/properties/name"
           x-field-uid: 1
-        metric_groups:
+        metric_tags:
           description: >-
-            A list of metric groups used to disaggregate flows.
-            A metric group that does not exist for a flow group MUST return
+            A list of metric tags used to disaggregate flows.
```
PrasenjitAdhikary, line 20: What will be the expected behavior if "metric tags" list is empty?

result/flow.yaml
anjan-keysight: I think this should always be in hex, (ideally with 0x prepended). The string format, at least, for egress tracking would be incorrect, because we can not assume DUT behaving correctly and there is no 'overlap' of non-tracked bits for the same field and we get unique string values for unique tracked bit combination. May be this will work for ingress tracking; but to keep response format consistent. we should use hex, always.

apratimmukherjee: Could not follow , if we have egress tracking set to track all 32 bits of a src ip ( lets assume it is theoritically possible)  , and return src ip in value as 1.1.1.1, 1.1.1.2 etc, ( with Rx for each ) --> How can there be 'overlap' of any bits ? Note I am only talking about use-cases where user has specified tracking on full field ( not on some custom offset or/and length in the packet or in a specific field ) .

result/flow.yaml
anjan-keysight: I think we should specify that the disaggregation would apply only for the specified "metric tags"; all configured tracking that is not specified in result, stats will be consolidated for them. So by default, empty "metric tags" will have no disaggregated stats.

result/flow.yaml
anjan-keysight: Tracking on full field, you will get unique strings; but to keep the response format consistent, I think, we should have hex format.

flow/flow.yaml
```diff
@@ -29,27 +29,47 @@ components:
...
             The size of the packets.
           $ref: './size.yaml#/components/schemas/Flow.Size'
-          x-field-uid: 3
+          x-field-uid: 5
         rate:
           description: |-
             The transmit rate of the packets.
           $ref: './rate.yaml#/components/schemas/Flow.Rate'
-          x-field-uid: 4
+          x-field-uid: 6
```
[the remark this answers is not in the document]
ashutshkumr: fixed

flow/flow.yaml
```diff
@@ -29,27 +29,47 @@ components:
...
         duration:
           description: |-
             The transmit duration of the packets.
           $ref: './duration.yaml#/components/schemas/Flow.Duration'
-          x-field-uid: 5
+          x-field-uid: 7
```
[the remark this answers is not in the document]
ashutshkumr: fixed

flow/flow.yaml
```diff
@@ -29,27 +29,47 @@ components:
...
         metrics:
           description: |-
             Flow metrics. 
           $ref: './metrics.yaml#/components/schemas/Flow.Metrics'
-          x-field-uid: 6
+          x-field-uid: 8
```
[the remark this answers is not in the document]
ashutshkumr: fixed

ashutshkumr:
> it would be easier if there was a `choice` to pick either an offset+length or reference it from predefined header field from the flow object

@parthpower I've added new example to demonstrate the same. If offset/length is not provided, it by default refers to the entire field.

```go
// Track egress IPv4 TOS precedence
egIpF1.Priority().Tos().Precedence().MetricTags().Add().SetName("egIpF1Tos")
```

apratimmukherjee (commented):
Clearing blocking comments based on current changes (esp. drilldown part and x-field uid changes)
Remaining items : How will future non packet content based tracking such as MPLS Flow Descriptor and track by size fit in , maybe in future , but in backward compatible manner. 
Will we continue to have hex returns and queries for fully tracked fields esp. for say MAC or IPv6 ?

PrasenjitAdhikary:
Expected result (shown here in tabular format for convenience):
| Flow | inVlanF1Id (2bit) | inIpF1Tos (full - 3bit) | egIpF1Tos (full - 3bit) | egIpF1Dst (2 bit) | FramesTx | FramesRx | Type                                                     | Comments           |
|------|-------------------|-------------------------|-------------------------|-------------------|----------|----------|----------------------------------------------------------|--------------------|
| f1   | 0x1               | 2                       |                         |                   | 50       | 50       | Aggregated stats for each combination of ingress fields. | missing in the example |
| f1   | 0x1               | 2                       | 5                      | 0x3               |          | 50       | Further drill-down for egress fields.                    |                    |
| f1   | 0x2               | 3                       |                         |                   | 50       | 50       | Aggregated stats for each combination of ingress fields  | missing in the example |
| f1   | 0x2               | 3                       | 6                       | 0x4               |          | 49       | Further drill-down for egress fields.                    |                    |
| f1   | 0x2               | 3                       | 7                       | 0x4               |          | 1        | Further drill-down for egress fields.                    |                    |
| f1   |                   |                         |                         |                   | 100      | 100      | Aggregated for the entire flow.                          |                    |

result/flow.yaml
```diff
@@ -45,47 +47,31 @@ components:
...
           type: string
-          default: ingress
           x-field-uid: 1
-          x-enum:
-            ingress:
-              x-field-uid: 1
-            egress:
-              x-field-uid: 2
-        ingress:
+        filters:
```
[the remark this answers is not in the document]
PrasenjitAdhikary: If two tags are specified (with some filter value as well), then the result will be filtered if matches any tag or both tags. There could be situations (a) each tag is associated with a different flow (b) both tags are associated with the same flow.

ankur-sheth (approved):

flow/flow.yaml
```diff
@@ -29,7 +29,27 @@ components:
...
+          description: |-
+            Ingress packet
+          type: array
+          minItems: 1
+          items:
+            $ref: './packet-headers/header.yaml#/components/schemas/Flow.Header'
+          x-field-uid: 8
+        egress_packet:
+          description: |-
+            Egress packets
```
PrasenjitAdhikary, line 47: At least, we need to clarify that in the Egress packet structure, header field value choices are not to be configured. Only attribute to be configured is "metric_tags"

result/flow.yaml
```diff
@@ -45,47 +47,31 @@ components:
...
-            If the items is empty then no disaggregation will be done. 
+            A list of filters that can be applied to the metric tag name.
+            By default all values will be included in the flow metric results.
           type: array
-          default: []
           items:
             type: string
-          x-constraint:
-          - "/components/schemas/Flow/properties/packet/../metric_group"
+            format: hex
```
PrasenjitAdhikary, line 69: Restriction on hex format should be removed. If the entire field is tracked, then the filter value is expected to be in the native format of that field.

result/flow.yaml
```diff
@@ -173,18 +185,98 @@ components:
...
+          x-field-uid: 2
+    Flow.MetricTag.Value:
+      description: |-
+        A container for metric tag value
+      type: object
+      properties:
+        choice:
+          description: Available formats for metric tag value
+          type: string
+          default: hex
```
PrasenjitAdhikary, line 271: For query, need an option where value type won't be specified. An 'auto-type might work - which will be translated to a value matching with the tag.

ashutshkumr (approved):

PR #313, opened 2023-03-27
Support of IMIX/Other packet size distribution

PrasenjitAdhikary:
Added support of IMIX/Custom IMIX/Quad Gaussian and other distributions in the model.
 
View proposed changes in [documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-frame-others/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config)

The following hierarchy is proposed for Frame Size:
	
```
Size:
  +-- Fixed
  +-- Increment
  +-- Random
  +-- Weight Pairs (including IMIX)
  |      +-- Predefined
  |      |      +-- IMIX
  |      |      ...  
  |      +-- Custom
  +-- Gaussian
  +-- <any new distribution to be added in future>
```

We have also discussed internally on the following hierarchy, but could not decide. Need feedback to finalize the hierarchy.
	  
```
Size:
  +-- Fixed
  +-- Increment
  +-- Random
  +-- Statistical Distribution
		  +-- Weight Pairs (including IMIX)
		  |      +-- Predefined
		  |      |      +-- IMIX
		  |      |      ...  
		  |      +-- Custom
		  +-- Gaussian
		  +-- <any new SD to be added in future>
```

Benefit:
 * Future proof – the addition of new statistically generated distributions can easily fit in the new group.

Drawback:
 * IMIX / Gaussian is becoming a more nested choice – not a top-level option, and easy to see at first glance. Need to expand the choice.

Also, need your comments to converge on:
1.	In weight pairs, the datatype for ‘weight’ is changed to float instead of an integer, considering the model will allow more granularity and fractional distribution for each frame size. However, our implementation is limited to integral value. Implication – if the fractional value is entered, Ixia-C implementation will round the value to the nearest ceiling and throw a warning on SetConfig for the implicit conversion. To avoid the warning, the user might update the configuration with an integral value.

2.	Instead of “Quad Gaussian” the distribution name is modified to “Gaussian” only. We have removed the hardcoded limit of the 4-gaussian curve, to make the model more open and future-proof. However, our model will be limited to supporting of maximum 4 Gaussian distributions in HW (software support is not available in USstream). Another reason to remove the “Quad” prefix, in the existing implementation, all four distributions are to be provided always, but the user can effectively disable up to 3 distributions by putting Zero weightage. So, distribution is not always “Quad Gaussian”. Instead, to achieve the same goal, we reduced the min-count (for the Gaussian curve) to 1. And max is implementation dependant.

4.	In Gaussian distribution, if attribute names should be continued with existing ixia-specific terms (center, width_at_half), or should be replaced with standard terminology (mean, standard deviation)?


flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+          x-field-uid: 3
+    Flow.SizeWeightPairs.Custom:
+      type: object
+      description: |-
+        Option allow to set a custom frame size distribution <size, weight> pairs.
+      properties:
+        size:
+          description: The size of the frame (in bytes) for this weight pair.
+          type: integer
+          minimum: 12
```
apratimmukherjee, line 148: Should the minimum frame size not be 64 ? What does 12 byte frame mean ?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+          x-enum:
+            predefined:
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        predefined:
+          type: string
+          default: imix
+          description: |-
+            Option allow to specify a predefined frame size distribution <size, weight> pairs (including IMIX distribution).
```
apratimmukherjee, line 104: I think we can start with "Specify pre-defined frame size ...

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
           type: integer
           default: 1518
           x-field-uid: 2
+    Flow.SizeWeightPairs:
+      type: object
+      description: |-
+        Custom frame size distribution, defined as <size, weight> pairs (including IMIX distribution).
```
apratimmukherjee, line 89: Maybe we should not add the IMIX distribution here since that is pre-defined and not custom ?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+            predefined:
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        predefined:
+          type: string
+          default: imix
+          description: |-
+            Option allow to specify a predefined frame size distribution <size, weight> pairs (including IMIX distribution).  
+            The available predefined distribution pairs are:
```
apratimmukherjee, line 105: Is there a source document for these pre-defined distributions if someone wants to check details or check if more have been added in future ?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+              x-field-uid: 9
+        custom:
+          type: array
+          items:
+            $ref: '#/components/schemas/Flow.SizeWeightPairs.Custom'
+          x-field-uid: 3
+    Flow.SizeWeightPairs.Custom:
+      type: object
+      description: |-
+        Option allow to set a custom frame size distribution <size, weight> pairs.
```
apratimmukherjee, line 143: Maybe just "Custom frame size ...   pair ( not pairs since this itself is one instance ) "

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+      properties:
+        size:
+          description: The size of the frame (in bytes) for this weight pair.
+          type: integer
```
apratimmukherjee, line 147: Maybe can be of format uint32 ( looks applicable for this field and supported by tools )

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+          minimum: 12
+          maximum: 65535
+          default: 64
+          x-field-uid: 1
+        weight:
+          description: A user-defined, 32-bit integer (weight) assigned to the corresponding frame size in this weight pair.
```
apratimmukherjee, line 153: Maybe good to specify higher weight means more packets or less packets of that size in the final generated stream of packets ?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+          default: 200.00
+          x-field-uid: 1
+        width_at_half:
+          description: |-
+            The width of the distribution curve at half of the curve weight, expressed as 
+            the difference between the minimum and maximum frame size in bytes.
+          type: number
+          format: double
+          minimum: 0.01
+          default: 100.00
```
apratimmukherjee, line 180: If max width really infinity ?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+          x-field-uid: 2
+        weight:
+          description: |-
+            The priority to give the distribution curve.
```
apratimmukherjee, line 184: Again, good to specify what it would mean for priority to be higher or lower? Same priority would mean all 4 flows would have same representation and high priority would mean that more of the packets would conform to the distribution with higher priority?

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+            cisco:
+              x-field-uid: 8
+            tolly:
+              x-field-uid: 9
+        custom:
+          type: array
+          items:
+            $ref: '#/components/schemas/Flow.SizeWeightPairs.Custom'
+          x-field-uid: 3
+    Flow.SizeWeightPairs.Custom:
```
apratimmukherjee, line 140: Interesting RFC https://www.rfc-editor.org/rfc/rfc6985.html : IMIX Genome: Specification of Variable Packet Sizes for Additional Testing : Both Ixia and Spirent in reference. Might not be relevant but looks to be a new standardized way to specify custom mix. More of a FYI than a comment.

flow/size.yaml
```diff
@@ -62,3 +83,107 @@ components:
...
+      type: object
+      description: |-
+        Option allow to set a custom frame size distribution <size, weight> pairs.
+      properties:
+        size:
```
apratimmukherjee, line 145: This raised a question that if we create a IMIX traffic and we see traffic drops , dont we need to have a way to track flows by packet size to see if packets of certain size are the ones getting dropped when running an IMIX test ? ( Relevant more for tracking PR )

ajbalogh (approved):
lgtm

flow/size.yaml
PrasenjitAdhikary: Agree. All default/min/max values are set according to IxNetwork, which also supports non-ethernet frames. Assuming OTG implementation is limited to an Ethernet frame, min should be 64. Some discrepancy is already present - increment & fixed frame are already accepting min up to 1. And as a test tool, do we restrict to 64 - or lower frame size that can be generated for -ve test scenario?

flow/size.yaml
PrasenjitAdhikary: Didn't find any RFC, but these mix are proposed by different test vendors (e.g. [ixia ref doc](http://www.ixiacom.com/pdfs/test_plans/agilent_journal_of_internet_test_methodologies.pdf) ).
All pre-defined options proposed in this model are based on IxNetwork.

flow/size.yaml
PrasenjitAdhikary: done

flow/size.yaml
PrasenjitAdhikary: done.

flow/size.yaml
PrasenjitAdhikary: done.

flow/size.yaml
PrasenjitAdhikary: done.

flow/size.yaml
```diff
@@ -62,3 +83,111 @@ components:
...
+        choice:
+          type: string
+          default: predefined
+          x-field-uid: 1
+          x-enum:
+            predefined:
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        predefined:
```
ashutshkumr, line 100: Few questions about predefined distribution:
- which ones are vendor neutral ?
- do we have a link or doc backing what each distribution looks like ?

IMHO, need to understand what value would adding predefined distribution add ? Since we can cover these using custom.

flow/size.yaml
```diff
@@ -62,3 +83,111 @@ components:
           type: integer
           default: 1518
           x-field-uid: 2
+    Flow.SizeWeightPairs:
+      type: object
+      description: |-
+        Frame size distribution, defined as <size, weight> pairs (including IMIX distribution).
```
ashutshkumr, line 89: Any comment on repeatability of randomness ?

flow/size.yaml
```diff
@@ -62,3 +83,111 @@ components:
...
+        weight:
+          description: |-
+            A user-defined, 32-bit integer (weight) assigned to the corresponding frame size in this weight pair. 
+            Higher weight means more packets.
+          type: integer
+          format: uint32
+          minimum: 1
+          default: 1
+          x-field-uid: 2
+    Flow.Size.Gaussian:
```
ashutshkumr, line 162: Any comment on repeatability of randomness ?

ashutshkumr:
Is there a way to document that if frame size is auto-adjusted, it might affect the distribution ?

e.g. eth/ip/tcp/payload may result in auto-adjusting 64B to higher value.

ashutshkumr:
The way I see it, following can be an alternate hierarchy:
- Fixed
- Increment
- StatisticalDistribution
   - WeightedPair
   - Gaussian
   - Predefined (don't really know if we need this - as it may be dependent on underlying header configuration)

ashutshkumr:
Moreover, `random` might need to be removed and inserted inside `Distribute` with following data structure: 
```
Random:
  repeatable: true
  weights: [(min, max, weight)]
```

This won't be a breaking hopefully as none of the implementations support random. But would suggest that we incorporate this if we do the restructuring.

flow/size.yaml
ashutshkumr: The reason for asking this is, let's assume we have a following distribution:
- 64B => 2
- 128B => 3

Possible sequences could be:
1. Exhaustive (Total packets 10): 64 64 64 64 128 128 128 128 128 128
2. Repeat after minimum exhaustion (Total packets: 10): 64 64 128 128 128 64 64 128 128 128
3. Uniformly distributed (Total packets: 3): 64 128 128
4. Non-uniform (Total packets 10): 64 128  64 64 64 128 128 128 128 128

flow/size.yaml
```diff
@@ -62,3 +83,111 @@ components:
...
+          format: uint32
+          minimum: 64
+          maximum: 65535
+          default: 64
+          x-field-uid: 1
+        weight:
+          description: |-
+            A user-defined, 32-bit integer (weight) assigned to the corresponding frame size in this weight pair. 
+            Higher weight means more packets.
+          type: integer
```
ashutshkumr, line 157: Shouldn't weight be float with minimum=0 to allow more distributions ?

Moreover, having more weight means packets will be proportionately higher. I think we need to document how the weights should be interpreted and what kind of distribution is expected (provided example on a separate comment)

flow/size.yaml
```diff
@@ -26,6 +30,23 @@ components:
         random:
           $ref: '#/components/schemas/Flow.SizeRandom'
           x-field-uid: 4
+        weight_pairs:
+          $ref: '#/components/schemas/Flow.SizeWeightPairs'
+          x-field-uid: 5
+        quad_gaussian:
```
ashutshkumr, line 36: do we need to be fix the size to 4 and call it quad ? shouldn't it just be generic with minItems=1 ?

I didn't quite get what we mean by "normalized to single distribution". To me it looks more like size-weight pairs, just that `fixed size` frame size has been replaced with range of frame sizes represented by `gaussian curve`.

flow/size.yaml
PrasenjitAdhikary: Done

flow/size.yaml
PrasenjitAdhikary: Explained with image reference.

flow/size.yaml
PrasenjitAdhikary: We should have a feature to track by frame size.

flow/size.yaml
PrasenjitAdhikary: Weight pair distribution only ensures the distribution ratio will be maintained over a fair number of packets. The sequence is not guaranteed.

flow/size.yaml
PrasenjitAdhikary: discussed in the meeting.

flow/size.yaml
PrasenjitAdhikary: done

flow/size.yaml
```diff
@@ -62,3 +83,111 @@ components:
...
+          type: integer
+          format: uint32
+          minimum: 1
+          default: 1
+          x-field-uid: 2
+    Flow.Size.Gaussian:
+      type: object
+      description: |-
+        Gaussian distribution.
+      properties:
```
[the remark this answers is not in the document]
PrasenjitAdhikary: explained with graphic.

flow/size.yaml
PrasenjitAdhikary: removed the fixed limit.

bortok (changes requested):
@PrasenjitAdhikary could you please reduce the proposed changes to only include those that are currently being requested by the OTG model consumers? We should keep the actual changes to those practically needed to avoid modeling something only because it can be supported by some implementations. This way, we adhere to vendor neutrality and avoid implementations that only partially overlap.

It is valuable to consider possible extensions like Gaussian distribution you described, so that we know the model can accommodate them if needed, but let's keep such options in theory until requested by the consumers.

PR #315, opened 2023-05-10
Support of IMIX packet size distribution

PrasenjitAdhikary:
Added support of IMIX/Custom IMIX distributions in the model.

View proposed changes in [documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-frame-imix/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config)

This model is a reduced version of the original [PR](https://github.com/open-traffic-generator/models/pull/313).
This new proposal is based on comments on the original [PR](https://github.com/open-traffic-generator/models/pull/313) and initial requirements as discussed on the 4th May 2023 Sprint demo meeting. There might be a need for additional options once the final requirement is available from Google.


flow/size.yaml
```diff
@@ -16,6 +16,10 @@ components:
               x-field-uid: 2
             random:
               x-field-uid: 3
+            weight_pairs:
+              x-field-uid: 4
+            quad_gaussian:
```
ajbalogh, line 21: Is this a placeholder or should it be removed as there is no property for this choice?

PrasenjitAdhikary: Missed this one - Removed.

ajbalogh (approved):

flow/size.yaml
```diff
@@ -62,3 +67,70 @@ components:
...
+            - IMIX (64:7, 570:4, and 1518:1)  
+            - IPSec IMIX (90:58.67, 92:2, 594:23.66 and 1418:15.67)  
+            - IPv6 IMIX (60:58.67, 496:2, 594:23.66 and 1518:15.67)  
+            - Standard IMIX (58:58.67, 62:2, 594:23.66 and 1518:15.67)  
+            - TCP IMIX (90:58.67, 92:2, 594:23.66 and 1518:15.67)  
+          x-field-uid: 2
+          x-enum:
+            imix:
+              x-field-uid: 1
+            ipsec_imix:
```
winstonliu1111, line 103: is this supported by some implementations?

flow/size.yaml
```diff
@@ -62,3 +67,70 @@ components:
...
+          description: The size of the frame (in bytes) for this weight pair.
+          type: integer
+          format: uint32
+          minimum: 12
+          maximum: 65535
+          default: 64
+          x-field-uid: 1
+        weight:
+          description: |-
+            Weight assigned to the corresponding frame size in this weight pair.
```
ankur-sheth, line 131: I am assuming that all the weights have to add up to 100?  If so, we should clearly call it out and also define the behavior when the weights are <100 or >100.  Or if it is implementation specific/un-defined, then explicitly state that.

flow/size.yaml
```diff
@@ -62,3 +67,70 @@ components:
...
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        predefined:
+          type: string
+          default: imix
+          description: |-
+            Specify predefined frame size distribution <size, weight> pairs (including IMIX distribution).  
+            The available predefined distribution pairs are:
+            - IMIX (64:7, 570:4, and 1518:1)
```
ankur-sheth, line 94: The weights of the IMIX distribution do not add up to 100, while the weights in all the other pre-defined distributions add up to 100.

ankur-sheth (changes requested):

flow/size.yaml
PrasenjitAdhikary: It is not necessary that all the weights have to add up to 100. For example, we may have weight pairs defined as (64,1) and (1500, 1000). It means, for every 1001 frames transmitted, 1 packet of 64 bytes, and the remaining 1000 packets of 1500 bytes.

flow/size.yaml
PrasenjitAdhikary: Predefined distributions are specified in %, but custom weight pairs are more flexible, and any distribution can be easily achieved. For example, if we have a sample size of 4M packets, and we know the packet count for each distinct frame size, the same can be easily translated to the custom pattern, where the sum of weights will be 4M.

flow/size.yaml
PrasenjitAdhikary: In IxNetwork, supported in IXHW and IxVM. In OTG, it will be supported in IXHW and Ixia-C SW ports. UHD support is under discussion.

ashutshkumr (approved):

winstonliu1111 (approved):

apratimmukherjee (approved):

flow/size.yaml
ankur-sheth: Ok.  No issues with not being % and not adding up to 100.  That is fine, just one minor nit - IMIX is pre-defined and not specified like the other pre-defined distributions.

PrasenjitAdhikary: All distributions are pre-defined, under the "predefined" category.  The user needs to select one of the available choices. So,  should we remove the word "Specify" from the description to read it like "Predefined frame size distribution ..." or add one more line in the description "IMIX is pre-defined and not specified like the other pre-defined distributions."? Could you please help with the exact change/wording on the nit?

ankur-sheth (approved):

Issue #319, opened 2023-05-25
Metrics response disaggregation based on Rx ports


anjan-keysight:
We require support for drill down of stats based on Rx ports when we support multi Rx port device (and raw) traffic. Currently we are aggregating port level stats and only segregate at flow and egress tracking level.

PR #320, opened 2023-05-25
Dev port disaggregation

anjan-keysight:
Support to address issue #319 (Metrics response disaggregation based on Rx ports).

This is Rx port level disaggregation of metrics response required for multi Rx port device (and raw) traffic. Also addressed is support for specifying Rx/Tx frames ratio.

Example code snippet
```
config.Ports().Add().SetName("p1").SetLocation("localhost:5555")
config.Ports().Add().SetName("p2").SetLocation("localhost:5555")
config.Ports().Add().SetName("p3").SetLocation("localhost:5555")
```

For control plane BGP sessions configured on P1, P2 and P3 with device names
p1d1peer1rrv4, p2d1peer1rrv4 and p3d1peer1rrv4 respectively.
Following is the data plane configuration

```
flow := config.Flows().Add()
flow.Metrics().SetEnable(true).PredefinedMetricTags().SetRxName(true)
```

We are expecting Tx and Rx packets to match in this example and hence the Tx Rx ratio is set to "one"
```
flow.Metrics().RxTxRatio().SetChoice("value")
flow.Duration().FixedPackets().SetPackets(1000)

flow.SetName("f1").TxRx().Device().
	SetTxNames([]string{"p1d1peer1rrv4"}).
	SetRxNames([]string{"p2d1peer1rrv4", "p3d1peer1rrv4"})

flowEth := flow.Packet().Add().Ethernet()
flowEth.Src().SetValue("00:00:01:01:01:01")
flowEth.Dst().SetChoice("auto")

flowIp := flow.Packet().Add().Ipv4()
flowIp.Src().SetValue("10.10.10.1")
flowIp.Dst().SetValues([]string{"20.20.20.1", "30.30.30.1"})

client.SetConfig(config)
```

Query for metrics results
```
client.GetFlowMetrics(nil, nil)

Flow      Frames Tx      Bytes Tx      Frames Rx     Bytes Rx      Tags
f1        1000           64000         1000          64000
f1                                     500           32000         rx_name: p2
f1                                     500           32000         rx_name: p3

```
Note: The “rx_name” would be in-built tag name for Rx ports and cannot be used for any egress tracking name.

### DIFF
--- a/flow/endpoint.yaml
+++ b/flow/endpoint.yaml
@@ -58,11 +58,11 @@ components:
           x-field-uid: 2
         rx_names:
           description: |-
-            Unique name of ports or lags that are intended receiving endpoints.
+            Unique name of ports or lags that are intended receive endpoints.
           type: array
           items:
             description: |-
-              The unique name of a port that is the intended receive port.
+              The unique name of a port or lag that is the intended receive port.
             type: string
           x-constraint:
           - '/components/schemas/Port/properties/name'
--- a/flow/endpoint.yaml
+++ b/flow/endpoint.yaml
@@ -58,7 +58,7 @@ components:
           x-field-uid: 2
         rx_names:
           description: |-
-            Unique names of ports that are intended receive ports.
+            Unique name of ports or lags that are intended receiving endpoints.
           type: array
           items:
             description: |-
--- a/flow/endpoint.yaml
+++ b/flow/endpoint.yaml
@@ -46,12 +46,28 @@ components:
         rx_name:
           description: |-
             The unique name of a port that is the intended receive port.
+          x-status:
+            status: deprecated
+            information: |-
+              This property is deprecated in favor of property rx_names
           type: string
           x-constraint:
           - '/components/schemas/Port/properties/name'
           - '/components/schemas/Lag/properties/name'
 
           x-field-uid: 2
+        rx_names:
+          description: |-
+            Unique names of ports that are intended receive ports.
+          type: array
+          items:
+            description: |-
+              The unique name of a port that is the intended receive port.
+            type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+          - '/components/schemas/Lag/properties/name'
+          x-field-uid: 3
     Flow.Router:
       description: >-
         A container for declaring a map of 1..n transmit devices
--- a/flow/metrics.yaml
+++ b/flow/metrics.yaml
@@ -19,6 +19,13 @@ components:
           type: boolean
           default: false
           x-field-uid: 2
+        rx_tx_ratio:
+          description: |-
+            Tx Rx multiplication factor, required for loss measurement.
+          type: number
+          format: float
+          default: 0
+          x-field-uid: 6
         timestamps:
           description: |-
             Enables additional flow metric first and last timestamps.
@@ -30,9 +37,9 @@ components:
             Latency metrics.
           $ref: '#/components/schemas/Flow.Latency.Metrics'
           x-field-uid: 4
-        rx_port_disaggregation:
+        rx_per_port:
           description: |-
-            Enables port level disaggregation with metrics tag name set as "rx_port_name".
+            Enables Rx port level disaggregation with metrics tag name set as "rx_port_name".
           type: boolean
           default: false
           x-field-uid: 5
--- a/flow/metrics.yaml
+++ b/flow/metrics.yaml
@@ -19,12 +19,10 @@ components:
           type: boolean
           default: false
           x-field-uid: 2
-        rx_tx_ratio:
+        tx_rx_replication_factor:
           description: |-
-            Tx Rx multiplication factor, required for loss measurement. Should be a positive, non-zero value.
-          type: number
-          format: float
-          default: 1
+            Tx Rx replication factor.
+          $ref: '#/components/schemas/Flow.TxRxReplication'
           x-field-uid: 6
         timestamps:
           description: |-
@@ -40,6 +38,7 @@ components:
         rx_per_port:
           description: |-
             Enables Rx port level disaggregation with metrics tag name set as "rx_port_name".
+            The Rx port names can be found under tagged_metrics tag names in flow metrics response.
           type: boolean
           default: false
           x-field-uid: 5
@@ -82,3 +81,44 @@ components:
               x-field-uid: 1
             cut_through:
               x-field-uid: 2
+    Flow.TxRxReplication:
+      description: |-
+        Tx Rx replication factor is required to determine the ratio of expected number of Rx packets across all Rx ports
+        to the Tx packets for the configured flow.
+      type: object
+      properties:
+        choice:
+          type: string
+          default: one
+          x-field-uid: 1
+          x-enum:
+            one:
+              x-field-uid: 1
+            rx_ports_count:
+              x-field-uid: 2
+            custom:
+              x-field-uid: 3
+        one:
+          $ref: '#/components/schemas/Flow.TxRxReplication.Ecmp'
+          x-field-uid: 2
+        rx_ports_count:
+          $ref: '#/components/schemas/Flow.TxRxReplication.RxCount'
+          x-field-uid: 3
+        custom:
+          description: |-
+            Should be a positive, non-zero value. We may need to specify a custom integer value for
+            loss calculation for cases when there are multiple destination addresses configured within one flow,
+            but DUT is configured to replicate only to a subset of Rx ports. For cases when Tx side generates two
+            packets from each source in 1:1 protection mode but only one of the two packets are received by the
+            Rx port, we may need to specify a fractional value instead.
+          type: number
+          format: float
+          default: 1.0
+          x-field-uid: 4
+    Flow.TxRxReplication.Ecmp:
+      description: |-
+        This is typically for cases where Rx packet count across all ports is expected to match the Tx packet count.
+    Flow.TxRxReplication.RxCount:
+      description: |-
+        This is for cases where one copy each Tx packet is received on all Rx ports and so the sum total of Rx packets
+        received across all Rx ports is a multiple of Rx port count and Tx packets.
--- a/flow/metrics.yaml
+++ b/flow/metrics.yaml
@@ -22,7 +22,7 @@ components:
         tx_rx_ratio:
           description: |-
             Tx Rx ratio.
-          $ref: '#/components/schemas/Flow.TxRxReplication'
+          $ref: '#/components/schemas/Flow.TxRxRatio'
           x-field-uid: 6
         timestamps:
           description: |-
@@ -81,10 +81,12 @@ components:
               x-field-uid: 1
             cut_through:
               x-field-uid: 2
-    Flow.TxRxReplication:
+    Flow.TxRxRatio:
       description: |-
-        Tx Rx ratio is the expected number of Rx packets across all Rx ports to the Tx packets
-        for the configured flow.
+        Tx Rx ratio is the ratio of expected number of Rx packets across all Rx ports to the Tx packets
+        for the configured flow. It is a factor by which the Tx packet count is multiplied to calculate
+        the sum of expected Rx packet count, across all Rx ports. This will be used to calculate loss
+        percentage of flow at aggregate level.
       type: object
       properties:
         choice:
@@ -99,10 +101,10 @@ components:
             custom:
               x-field-uid: 3
         one:
-          $ref: '#/components/schemas/Flow.TxRxReplication.Ecmp'
+          $ref: '#/components/schemas/Flow.TxRxRatio.Ecmp'
           x-field-uid: 2
         rx_count:
-          $ref: '#/components/schemas/Flow.TxRxReplication.RxCount'
+          $ref: '#/components/schemas/Flow.TxRxRatio.RxCount'
           x-field-uid: 3
         custom:
           description: |-
@@ -115,10 +117,10 @@ components:
           format: float
           default: 1.0
           x-field-uid: 4
-    Flow.TxRxReplication.Ecmp:
+    Flow.TxRxRatio.Ecmp:
       description: |-
         This is for cases where Rx packet count across all ports is expected to match the Tx packet count.
-    Flow.TxRxReplication.RxCount:
+    Flow.TxRxRatio.RxCount:
       description: |-
         This is for cases where one copy each Tx packet is received on all Rx ports and so the sum total of Rx packets
         received across all Rx ports is a multiple of Rx port count and Tx packets.
--- a/flow/metrics.yaml
+++ b/flow/metrics.yaml
@@ -19,9 +19,9 @@ components:
           type: boolean
           default: false
           x-field-uid: 2
-        tx_rx_replication_factor:
+        tx_rx_ratio:
           description: |-
-            Tx Rx replication factor.
+            Tx Rx ratio.
           $ref: '#/components/schemas/Flow.TxRxReplication'
           x-field-uid: 6
         timestamps:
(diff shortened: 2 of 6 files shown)


anjan-keysight:
The response format has been captured [here](https://keysighttech.sharepoint.com/:w:/r/sites/Athena-G/Shared%20Documents/Egress_Tracking/Rx%20port%20disaggregation%20stats.docx?d=w63c1ad1e14b54d4f820fbb62be9ba23f&csf=1&web=1&e=n2cWdJ).

parthpower:
@anjan-keysight 
> The response format has been captured [here](https://keysighttech.sharepoint.com/:w:/r/sites/Athena-G/Shared%20Documents/Egress_Tracking/Rx%20port%20disaggregation%20stats.docx?d=w63c1ad1e14b54d4f820fbb62be9ba23f&csf=1&web=1&e=n2cWdJ).

linked document is not accessible to public

anjan-keysight:
@parthpower I have shared the document.

anjan-keysight:
[Rx port disaggregation stats.docx](https://github.com/open-traffic-generator/models/files/11611081/Rx.port.disaggregation.stats.docx)

flow/metrics.yaml
```diff
@@ -19,6 +19,13 @@ components:
           type: boolean
           default: false
           x-field-uid: 2
+        rx_tx_ratio:
+          description: |-
+            Tx Rx multiplication factor, required for loss measurement. Should be a positive, non-zero value.
```
apratimmukherjee, line 24: It might be more intuitive to 
a) Call it tx_rx_replication_factor
b) Have a set of enums : 
    one
    number_of_rx_ports
    custom 
         Here , user can give a float if needed 
   The default being 'one'
   No of rx_ports means for replicating use-case of multicast , normally replication factor can be auto calculated and be same as number of rx ports.

apratimmukherjee, line 24: Also, a more detailed explanation here would be useful, especially detailing the scenario when Rx packets = Tx packets * number of Rx ports for multicast or broadcast traffic.  
Finally , we could give some  explanation of scenarios when custom and non-integral values are needed 
e.g. there are multiple dst IPs configured within one flow , but DUT is configured to replicate only to a subset of Rx ports.
Or Tx side generates two packets from each source in 1:1 protection  but only one of the two packets are received by Rx ports ( so replication factor becomes 0.5 and not an integral number )

flow/metrics.yaml
```diff
@@ -30,6 +37,12 @@ components:
             Latency metrics.
           $ref: '#/components/schemas/Flow.Latency.Metrics'
           x-field-uid: 4
+        rx_per_port:
+          description: |-
+            Enables Rx port level disaggregation with metrics tag name set as "rx_port_name".
```
apratimmukherjee, line 42: ... metrics tag name automatically set by implementation as "rx_port_name"  in the flow metric_response ( use exact attribute names from metric response )

apratimmukherjee: Also , do we need updates in otg gnmi model to specify that response can be filtered for specific port by specifing the filter for rx_port_name:<port name>  in this exact format . Otherwise, how will gnmi user know what filter to use or expect which has to be parsed to set expectation in tests when checking stats using gnmi ( or otg metrics)

PrasenjitAdhikary: For gNMI query path: “/flows/flow[name=f1]/”
expected gNMI response (for the metric response example in PR description):

```
"updates": [
  {
	"Path": "flows/flow[name=f1]",
	"values": {
	  "flows/flow": {
		"open-traffic-generator-flow:name": "f1",
		"open-traffic-generator-flow:state": {                     // Contains the aggregated per-flow stats
		  "counters": {
			"in-pkts": "1000",
			"in-octets": "8000",
			"out-pkts": "1000",
			"out-octets": "8000"
		  },
		  "in-frame-rate": "QKAAAA==",
		  "name": "f1",
		  "out-frame-rate": "QCAAAA==",
		  "transmit": true
		},
		"open-traffic-generator-flow:tag-metrics": {              // Contains the disaggregated per-flow stats
		  "tag-metric": [
			{
			  "name-value": "rx_port_name=p2",
			  "state": {
				"counters": {
				  "in-octets": "4000",
				  "in-pkts": "500"
				},
				"name-value": "rx_port_name=p2",
				"tags": [
				  {
					"tag-name": "rx_port_name",
					"tag-value": {
					  "value-as-string": "p2",
					  "value-type": "STRING"
					}
				  }
				]
			  }
			},
			{
			  "name-value": "rx_port_name=p3",
			  "state": {
				"counters": {
				  "in-octets": "4000",
				  "in-pkts": "500"
				},
				"name-value": "rx_port_name=p3",
				"tags": [
				  {
					"tag-name": "rx_port_name",
					"tag-value": {
					  "value-as-string": "p3",
					  "value-type": "STRING"
					}
				  }
				]
			  }
			}
		  ]
		}
	  }
	}
  }
]
```

flow/metrics.yaml
anjan-keysight: Update the property name and made it into an enum.

anjan-keysight: Descriptions also updated for each enum value.

flow/metrics.yaml
anjan-keysight: Description modified to reflect Rx port name location in tagged response.

flow/metrics.yaml
```diff
@@ -69,3 +81,44 @@ components:
...
+            custom:
+              x-field-uid: 3
+        one:
+          $ref: '#/components/schemas/Flow.TxRxReplication.Ecmp'
+          x-field-uid: 2
+        rx_ports_count:
+          $ref: '#/components/schemas/Flow.TxRxReplication.RxCount'
+          x-field-uid: 3
+        custom:
+          description: |-
```
ankur-sheth, line 108: I'm not able to fully follow this description and what is the use-case for this.  Also refrain from "we may need to specify", instead state it as "A custom integer value (>1) can be specified for loss calculations in scenarios where there are multiple ..."

flow/metrics.yaml
```diff
@@ -69,3 +81,44 @@ components:
...
+            but DUT is configured to replicate only to a subset of Rx ports. For cases when Tx side generates two
+            packets from each source in 1:1 protection mode but only one of the two packets are received by the
+            Rx port, we may need to specify a fractional value instead.
+          type: number
+          format: float
+          default: 1.0
+          x-field-uid: 4
+    Flow.TxRxReplication.Ecmp:
+      description: |-
+        This is typically for cases where Rx packet count across all ports is expected to match the Tx packet count.
```
ankur-sheth, line 120: Is this the case were:
 sum(Rx packets across all ports) = sum(Tx packets across all ports)

flow/metrics.yaml
```diff
@@ -19,6 +19,11 @@ components:
           type: boolean
           default: false
           x-field-uid: 2
+        tx_rx_replication_factor:
```
ankur-sheth, line 22: I didn't follow what the user accomplishes by setting this configuration?  In the example provided, I don't see how it got used and how it impacted the statistics that were gathered and displayed.

ankur-sheth (changes requested):

bortok (commented):
I have a more general question. A device's `ethernet` can have a `connection` to a `port`, `lag` or `vxlan` tunnel. As a result, a flow that is received by multiple devices can come through all of these types of connections. Shall we disaggregate Rx metrics accordingly, per `lag` or `vxlan` when configured as such, instead of port-level disaggregation?

flow/metrics.yaml
```diff
@@ -30,6 +35,13 @@ components:
             Latency metrics.
           $ref: '#/components/schemas/Flow.Latency.Metrics'
           x-field-uid: 4
+        rx_per_port:
```
bortok, line 38: should we also have `rx_per_lag` and `rx_per_vxlan`? Or maybe if we want to support those in future, shall we do `rx_per_connection`?

flow/metrics.yaml
```diff
@@ -69,3 +81,44 @@ components:
               x-field-uid: 1
             cut_through:
               x-field-uid: 2
+    Flow.TxRxReplication:
```
bortok, line 84: Wouldn't `Replication` imply additional copies? Meaning, if we have `Replication = 1` it means we expect to receive twice more packets that we sent? Maybe we could use `TxRxRatio` instead?

flow/metrics.yaml
```diff
@@ -69,3 +81,44 @@ components:
...
+      description: |-
+        Tx Rx replication factor is required to determine the ratio of expected number of Rx packets across all Rx ports
+        to the Tx packets for the configured flow.
+      type: object
+      properties:
+        choice:
+          type: string
+          default: one
```
bortok, line 92: `one` makes much more sense for `TxRxRatio` than for `TxRxReplication`

flow/metrics.yaml
```diff
@@ -69,3 +81,44 @@ components:
...
+          x-field-uid: 1
+          x-enum:
+            one:
+              x-field-uid: 1
+            rx_ports_count:
```
bortok, line 97: If my Rx is a `lag` with 2 ports, and I'm broadcasting, `rx_ports_count` would mean I shall expect twice more packets on that `lag` than `tx` value. But that is wrong. So we should have `rx_connection_count` instead to count `lag` as one.

bortok: or just `rx_count`

bortok (changes requested):

flow/metrics.yaml
anjan-keysight: I have updated the description based on suggestion. Regarding how the Custom enum is used refer to the mail with Alex's example.

flow/metrics.yaml
anjan-keysight: Yes

flow/metrics.yaml
anjan-keysight: I have updated the example (how this is used). Also refer to mail reply with Alex's example.

flow/metrics.yaml
anjan-keysight: Currently we do not have support for drill down stats for lag or vxlan in TE. So we will support ports level disaggregation for now. But the model is very extensible and when we have the underlying support for those, we can extend the model without any breaking change.

flow/metrics.yaml
anjan-keysight: Yes, have updated to TxRxRatio.

flow/metrics.yaml
anjan-keysight: I have renamed it to just rx_count.

flow/metrics.yaml
bortok: should we change the reference as well from `$ref: '#/components/schemas/Flow.TxRxReplication'` to `$ref: '#/components/schemas/Flow.TxRxRatio'`?

flow/metrics.yaml
bortok: could you give an example of how the model would be extended? it would be good to see how that would be aligned with `rx_per_port` parameter. i'm thinking maybe we should use `choice` here and set it as `port` first and then extend to `lag` and `vxlan` later?

flow/metrics.yaml
anjan-keysight: Updated.

flow/metrics.yaml
anjan-keysight: I have restructured this part, now all additional tracking will come under "predefined_metric_tags" and we can add new properties under this. Current property "rx_name" will apply for both raw port and lag.

bortok (approved):

ajbalogh (approved):

ankur-sheth (approved):

flow/metrics.yaml
```diff
@@ -69,3 +79,50 @@ components:
...
+            but DUT is configured to replicate only to a subset of Rx ports. For cases when Tx side generates two
+            packets from each source in 1:1 protection mode but only one of the two packets are received by the
+            Rx port, we may need to specify a fractional value instead.
+          type: number
+          format: float
+          default: 1.0
+          x-field-uid: 3
+    Flow.RxTxRatio.RxCount:
+      description: |-
+        This is for cases where one copy each Tx packet is received on all Rx ports and so the sum total of Rx packets
```
apratimmukherjee, line 127: one copy of

apratimmukherjee (approved):
One minor typo.

flow/endpoint.yaml
```diff
@@ -46,12 +46,28 @@ components:
...
           - '/components/schemas/Lag/properties/name'
 
           x-field-uid: 2
+        rx_names:
+          description: |-
+            Unique names of ports that are intended receive ports.
+          type: array
+          items:
+            description: |-
+              The unique name of a port that is the intended receive port.
```
PrasenjitAdhikary, line 65: Should we correct 'port' to 'port or lag'?

flow/metrics.yaml
```diff
@@ -69,3 +79,50 @@ components:
...
+            rx_count:
+              x-field-uid: 1
+            value:
+              x-field-uid: 2
+        rx_count:
+          $ref: '#/components/schemas/Flow.RxTxRatio.RxCount'
+          x-field-uid: 2
+        value:
+          description: |-
+            Should be a positive, non-zero value. A custom integer value (>1) can be specified for
```
PrasenjitAdhikary, line 116: We might need to explain the scenario with the default value (i.e. == 1)

PrasenjitAdhikary (approved):
LGTM

flow/endpoint.yaml
jasdeep-hundal: Not seeing the updated change?

jasdeep-hundal (commented):
I'm still thinking about GNMI outputs a bit here, but I'm wondering if we should just use the port name directly as the tag ("<port>" vs. "rx_port_name=<port>".) (It should be a unique name in the config, but I'm also not sure if there are any other future 'built-in' ones that could not be specified this way.)

Unrelated clarification question: These tags will be included in disaggregated GNMI stats in the same way/alongside the user-defined ones correct?

flow/metrics.yaml
```diff
@@ -69,3 +79,51 @@ components:
...
             cut_through:
               x-field-uid: 2
+    Flow.Predefined.Tags:
+      description: |-
+        List of predefined flow tracking options, outside packet fields, that can be enabled.
+      type: object
+      properties:
+        rx_name:
+          description: |-
+            Enables Rx port or lag level disaggregation with predefined metrics tag name set as "rx_name".
```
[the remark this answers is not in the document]
anjan-keysight: Both otg and gnmi tag name is going to be rx_name (rx_port_name was the earlier proposal). The rx_name is generic tag name, referring to both port and lag as rx endpoints.
So if we have a flow configured from lag L1 to lag L2 and a separate port P5, the metrics response will look like:
F1:    L1  =>   L2, P5

Flow     Frames Tx     Bytes Tx     Frames Rx    Bytes Rx     Tags
f1         1000             64000        1000             64000
f1		                                   400               25600        rx_name: p5
f1		                                   600               38400        rx_name: l2

There can be other built-in tag names in future, like "mem_name" (as an example), if we want to disaggregate based on lag member ports also.
And yes, these tag names will be alongside user defined ones.

flow/metrics.yaml
```diff
@@ -69,3 +79,51 @@ components:
...
+            all ports is expected to match the Tx packet count. A custom integer value (>1) can be specified for
+            loss calculation for cases when there are multiple destination addresses configured within one flow,
+            but DUT is configured to replicate only to a subset of Rx ports. For cases when Tx side generates two
+            packets from each source in 1:1 protection mode but only one of the two packets are received by the
+            Rx port, we may need to specify a fractional value instead.
+          type: number
+          format: float
+          default: 1.0
+          x-field-uid: 3
+    Flow.RxTxRatio.RxCount:
```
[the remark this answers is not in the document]
anjan-keysight: User can use RxCount, so even if the rx port count changes in config, user does not have to update it.
But I leave it to you, if you feel otherwise.

PR #341, opened 2023-11-07
Support for BGP Update Replay

apratimmukherjee:
**Redocly View:** 
[ReDoc Interactive Demo (redocly.github.io)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/bgp_route_replay/artifacts/openapi.yaml&nocors#tag/Configuration) 
( New optional object at : set_config/devices/bgp/ipv[46]_interfaces/peers/replay_updates ) 

**Requirement:**
Provide a way for importing a list of updates available in a proprietary format into OTG format so that the stream of BGP Updates can be replayed to a Device Under Test which is implementing this model.
This stream of BGP Updates , which can include both Route Advertisements and Withdraws should be sent to the Device under Test in same order as supplied by user.

**Proposal:**
 What format to store the BGP Updates in for this neutral OTG format? 
A) as array of raw bytes
The proposal is to keep the model as simple as possible without any assumptions of format of attributes or TLVs being correct or incorrect.
Due to this , the proposal is to simply store the Updates as a list , with each element of the array being the array of bytes after the mandatory 19 byte header containing 16 Marker bytes , 2 bytes containing the Length and the 1 byte fixed Type containing value for BGP_UPDATE in it. 

B) In parsed human-readable format for user to be able to directly create it from test program or modify imported routes easily as well as to  ease the ability to import proprietary format Update stream to within the OTG conformant configuration to be consumed by test tools for Update Replay. 

When to perform conversion ?
In initial delivery , we plan to target the initial use-case for this when used from a test program written in go.
a) Create a OTG config with BGP peers configured. 
It is expected that BGP Capabilities will be configured based on Updates that are planned to be imported. If the corresponding capabilities , AS number , type of peer etc. are not configured to match the Updates being imported OTG implementation will still advertise the updates. The Device Under Test will reject updates if corresponding Capabilities are not enabled for the configured BGP peer as well as on the Device Under Test.

b) In the test program, import the Updates available in some proprietary format to the Structured Format ( or hex format ) as defined in below model.

c) Next, the usual workflow will be followed i.e. SetConfig and StartProtocols to advertise the imported Updates to the Device Under Test.

**Example usage: Create Replay Updates by manually specifying Structured Updates in the UpdateReplay list** 
```
/* Not including snippet upto creation of BGP Peer ( with AS type/AS number / Rtr ID etc. matching Updates shown below*/
updateReplayBlock := peer.ReplayUpdates().StructuredPdus()
adv1 := updateReplayBlock.Updates().Add()
adv.PathAttributes().
    SetOrigin(gosnappi.BgpAttributesOrigin.IGP)

adv1.PathAttributes().
	AsPath().FourByteAsPath().Segments().Add().
	SetType(gosnappi.BgpAttributesFourByteAsPathSegmentType.AS_SEQ).
    SetAsNumbers([]uint32{70000, 80000 })

adv1.PathAttributes().
	LocalPreference().SetValue(102)

adv1.PathAttributes().
	SetIncludeAtomicAggregator(true)

adv1.PathAttributes().
	Aggregator().
    SetFourByteAs(70001).
    SetIpv4Address("1.11.1.11")

adv1.PathAttributes().
	Community().Add().
	NoLlgr()

adv1.PathAttributes().
	OriginatorId().
    SetValue("11.11.11.11") //Normally a router ID but carried as integer

/* If IP matches that of DUT , we get a Cluster ID loop issue, so keep them distinct */
adv1.PathAttributes().
	SetClusterIds([]string{"1.1.1.22", "1.1.1.23"})

/* One Bandwidth extended community */
adv1.PathAttributes().
	ExtendedCommunities().Add().
	NonTransitive2OctetAsType().
	LinkBandwidthSubtype().
    SetGlobal2ByteAs(1112).
    SetBandwidth(10000.0)

/* 	Example of other attributes that can be encoded as hex bytes to be appended to the Update.
	Multiple such attributes can be added. */
adv1.PathAttributes().
	OtherAttributes().Add().
	SetFlagOptional(true).
    SetFlagTransitive(true).
    SetRawValue("10001AB0000432A")
				
/* Add reachable routes as MP_REACH .The NEXT_HOP is encoded inside the MP_REACH itself and not needed as a separate attribute*/
adv1.PathAttributes().MpReach().
	NextHop().
	SetIpv4("1.1.1.2")
ipv4_unicast_routes_adv := adv1.PathAttributes().MpReach().Ipv4Unicast()
ipv4_unicast_routes_adv.Add().
	SetAddress("100.100.1.0).SetPrefix(24)
ipv4_unicast_routes_adv.Add().
	SetAddress("100.100.4.0).SetPrefix(24)
ipv4_unicast_routes_adv.Add().
	SetAddress("100.101.0.0).SetPrefix(16)
        
/* Withdraw second route after gap of 5s */
adv2 := updateReplayBlock.Updates().Add().
			SetTimeGap(1000)
ipv4_unicast_routes_withdraw := adv2.PathAttributes().MpUnreach().Ipv4Unicast()
ipv4_unicast_routes_withdraw.Add().
	SetAddress("100.100.4.0).SetPrefix(24)                
```
**Example usage: Snippet for traditional REACH/UNREAC_NLRI withdraw advertise and withdraw ( only the routes , Path attributes will be configured in same way as in above example)** 
```
updateReplayBlock := peer.ReplayUpdates().StructuredPdus()
adv1 := updateReplayBlock.Updates().Add()
.... // Add various Path Attributes
adv1.TraditionalReachNlris().Add().
	SetAddress("100.100.1.0).SetPrefix(24)
adv1.TraditionalReachNlris().Add().
	SetAddress("100.100.2.0).SetPrefix(24)
adv2 := updateReplayBlock.Updates().Add().
			SetTimeGap(1000)
adv2.TraditionalUnreachNlris().Add().
	SetAddress("100.100.1.0).SetPrefix(24)	
```

**Example usage: Create Replay Updates by manually specifying raw bytes per updates in the UpdateReplay list 
    e.g. by using Copy As Hex Stream for Update contents and pasting the bytes from Path Attributes in a BGP Update/Withdraw in a pcap** 
```
   updateReplayBlock := d1BgpIpv4Interface1Peer1.ReplayUpdates().RawBytes()
        adv1 := updateReplayBlock.Updates().Add()
        adv1.SetUpdateBytes("4001010040020040050400000000c00804ffffff02800e16000149046401000100600000000100000064c8010001d0170010000f000c8000090001060000000100ff")

        /* Withdraw first route after gap of 100ms */
        /* UNREACH_NLRI only needs the prefix keys and attributes are normally not needed */
        adv2 := updateReplayBlock.Updates().Add().
                SetTimeGap(100)
        adv2.SetUpdateBytes("800f1b0001010000000118c80100000"+
                            "0000118c801010000000118c80102")


```
**Future enhancements:**
  These can be addressed via additional PRs in incremental manner as needed after receiving feedback from end-users.
  a) Add support for SEGMENT_ROUTING attributes in Path Attributes as well additional options for MP_REACH/UNREACH contents. 
  b)  Additional controls if needed e.g. modify next hop to match peer. 
  c)  Automatically identify and enable BGP capabilities if needed. .
  d) Add support for importing routes for replay from a pcap [ Maybe where user specifies specific BGP stream from which Updates need to be imported ]


device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,31 @@
+components:
+  schemas:
+    Bgp.UpdateReplay:
+      description: >-
+        Ordered BGP Updates ( including both Advertise and Withdraws ) to be sent to the peer after the BGP session is established.
```
kingshukdev, line 5: "to be sent in the order given in the input"

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,31 @@
...
+      type: object
+      properties:
+        updates:
+          description: >-
+            Array of ordered BGP Updates ( including both Advertise and Withdraws ) to be sent to the peer after the BGP session is established.
```
kingshukdev, line 10: Previous comment applies here also

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,31 @@
...
+    Bgp.OneUpdateReplay:
+      description: >-
+        Specification of one BGP Update to be sent to the BGP peer.        
+      type: object
+      properties:
+        update_bytes:    
+          description: >-
+            Bytes to be sent to peer after the BGP Update Header. The Update Header will always have the initial 16 bytes containing 
+            Marker bytes, 2 bytes containing the Length and 1 byte containing the Type.The string MUST contain sequence of valid hex bytes.
+            The bytes specified should be appended to  the Update message to be sent to the peer after the fixed 19 bytes above.
```
kingshukdev, line 24: Remove extra space after "append to".
"fixed 19 bytes above" ---> "fixed 19 bytes described above"

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,31 @@
...
+        Specification of one BGP Update to be sent to the BGP peer.        
+      type: object
+      properties:
+        update_bytes:    
+          description: >-
+            Bytes to be sent to peer after the BGP Update Header. The Update Header will always have the initial 16 bytes containing 
+            Marker bytes, 2 bytes containing the Length and 1 byte containing the Type.The string MUST contain sequence of valid hex bytes.
+            The bytes specified should be appended to  the Update message to be sent to the peer after the fixed 19 bytes above. 
+            The maximum length of 4077 is obtained from the maximum length allowed for a BGP message in RFC4271 which is 4096 minus mandatory 
+            19 bytes described above.
```
kingshukdev, line 26: Will suggest to add the following sentence :- 
"This byte stream can be of any length from <Apratim, give the min length> to 4077"

artifacts/otg.proto
```diff
@@ -3889,6 +3893,28 @@ message BgpGracefulRestart {
...
+}
+
+// Specification of one BGP Update to be sent to the BGP peer.
+message BgpOneUpdateReplay {
+
+  // Bytes to be sent to peer after the BGP Update Header. The Update Header will always
+  // have the initial 16 bytes containing  Marker bytes, 2 bytes containing the Length
+  // and 1 byte containing the Type.The string MUST contain sequence of valid hex bytes.
+  // The bytes specified should be appended to  the Update message to be sent to the peer
+  // after the fixed 19 bytes above.  The maximum length of 4077 is obtained from the
```
kingshukdev, line 3912: Since we are tacking about maximum, we should talk about minimum also.

artifacts/openapi.yaml
```diff
@@ -5381,6 +5386,31 @@ components:
...
+            $ref: '#/components/schemas/Bgp.OneUpdateReplay'
+          x-field-uid: 1
+    Bgp.OneUpdateReplay:
+      description: "Specification of one BGP Update to be sent to the BGP peer.  \
+        \      "
+      type: object
+      properties:
+        update_bytes:
+          description: |-
+            Bytes to be sent to peer after the BGP Update Header. The Update Header will always have the initial 16 bytes containing  Marker bytes, 2 bytes containing the Length and 1 byte containing the Type.The string MUST contain sequence of valid hex bytes. The bytes specified should be appended to  the Update message to be sent to the peer after the fixed 19 bytes above.  The maximum length of 4077 is obtained from the maximum length allowed for a BGP message in RFC4271 which is 4096 minus mandatory  19 bytes described above.
```
kingshukdev, line 5408: While talking about the maximum, talk about the minimum also.

kingshukdev (commented):
I am done with my part.

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Done.

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Done.

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Done.

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Done,

artifacts/otg.proto
apratimmukherjee: I specifically mentioned the logic for maximum since without this 4077 might appear to be an arbitrary max. Either way, I have added that the minimum length of the input byte stream should be 1 as per previous comment since we are allowing the user to send whatever he/she wants upto max length.

artifacts/openapi.yaml
apratimmukherjee: I specifically mentioned the logic for maximum since without this 4077 might appear to be an arbitrary max. Either way, I have added that the minimum length of the input byte stream should be 1 as per previous comment since we are allowing the user to send whatever he/she wants upto max length.

SuryyaKrJana (approved):
For the first delivery looks fine,

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,33 @@
+components:
+  schemas:
+    Bgp.UpdateReplay:
+      description: >-
+        Ordered BGP Updates ( including both Advertise and Withdraws ) to be sent in the order given in the input to the peer after the BGP session is established.
+      type: object
+      properties:
+        updates:
```
PrasenjitAdhikary, line 8: Is there any possibility of any other type(s) other than `updates` under `Bgp.UpdateReplay`? Otherwise, we might consider reducing one nesting level.

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,33 @@
...
+    Bgp.OneUpdateReplay:
+      description: >-
+        Specification of one BGP Update to be sent to the BGP peer.        
+      type: object
+      properties:
+        update_bytes:    
+          description: >-
+            Bytes specified in hex format to be sent to peer after the BGP Update Header. The Update Header will always have the 
+            initial 16 bytes containing Marker bytes, 2 bytes containing the Length and 1 byte containing the Type.The string MUST 
+            contain sequence of valid hex bytes.The bytes specified in hex format should be appended to the Update message to be sent
```
PrasenjitAdhikary, line 24: minor nit: space is missing before the new line.

device/bgp/bgpupdatereplay.yaml
```diff
@@ -0,0 +1,33 @@
+components:
+  schemas:
+    Bgp.UpdateReplay:
+      description: >-
+        Ordered BGP Updates ( including both Advertise and Withdraws ) to be sent in the order given in the input to the peer after the BGP session is established.
```
PrasenjitAdhikary, line 5: Do we need any control/timing gap between the two updates? Similarly, do we need any trigger event to replay the updates again or it will be done on restart of the protocol?

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Indeed I have kept it for future controls needed if any e.g. providing a name to the object to be able to use that as a key on some future TBD action. (e.g. replay by user trigger )

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Yes , this is kept in a object to allow per update controls such as inter update gap in between the list of updates. Not planning to add it all in one phase but gradually as needed once the basic feature is working.

PrasenjitAdhikary (approved):
Looks good to me. There are some questions which may not be applicable / you might already thought.

device/bgp/bgpupdatereplay.yaml
apratimmukherjee: Done.

jasdeep-hundal:
Some comments:
- I would have expected there to be a human-readable intermediate form here, instead of the raw byte representation.
- Does this possibly make more sense to just include the replay functionality as a control action? 
- Can we add a way to optionally time these updates, instead of just replaying back-to-back?

apratimmukherjee:
I have updated the model to allow support structured BGP updates in update replay . Also has option for inter update gap to be configurable. As of now , have kept it as part of SetConfig itself and can later add a secondary path via control action. 
In comments have added snippets of .proto as [.proto .... ] which I have kept temporarily to enable easier review and can be removed later. 
I will add code snippets after generating a dev snappi build from this PR branch in a couple of days to add example for structured updates using modified proposed model.

apratimmukherjee:
> I have updated the model to allow support structured BGP updates in update replay . Also has option for inter update gap to be configurable. As of now , have kept it as part of SetConfig itself and can later add a secondary path via control action. In comments have added snippets of .proto as [.proto .... ] which I have kept temporarily to enable easier review and can be removed later. I will add code snippets after generating a dev snappi build from this PR branch in a couple of days to add example for structured updates using modified proposed model.

Code snippets are now added for import via util library , or creating either structured or raw BGP updates for use in replay format .
 @jasdeep-hundal : Please check if this is matching your needs and please provide your comments. 
I will remove [.proto xxxx] parts from descriptions once review is done which I have kept for now to make review w.r.t. proto definition easier.

jasdeep-hundal:
@apratimmukherjee - Some comments:
- Thanks for catching the traditional vs. MP NLRI issue. Following on from that, it is possible to have both in an Update message correct? (From the RFC I think that as long as the prefixes don't overlap it should be a valid message.) Do we just want to just have two versions of the reach/unreach NLRIs? (Happy to hear other thoughts here, I do think surfacing the MP NLRIs at this level is the correct decision for clarity.)
- I'm still hesitant on this not being a control action from the start. (I see the comment 'Useful for non-golang users who cannot directly consume the updateReplayService go library', but does anything block that library usage for golang users if we just start with control actions?)
- 'structured_updates' definition overall looks good to me.

apratimmukherjee:
> @apratimmukherjee - Some comments:
> 
> * Thanks for catching the traditional vs. MP NLRI issue. Following on from that, it is possible to have both in an Update message correct? (From the RFC I think that as long as the prefixes don't overlap it should be a valid message.) Do we just want to just have two versions of the reach/unreach NLRIs? (Happy to hear other thoughts here, I do think surfacing the MP NLRIs at this level is the correct decision for clarity.)
> * I'm still hesitant on this not being a control action from the start. (I see the comment 'Useful for non-golang users who cannot directly consume the updateReplayService go library', but does anything block that library usage for golang users if we just start with control actions?)
> * 'structured_updates' definition overall looks good to me.
@jasdeep-hundal : I have updated the definition to separate MP_REACH / MP_UNREACH inside the Attributes and additional Traditional NLRIs at the Update level . 
Regarding control actions , few points I would like to understand and discuss in more depth is regarding triggers normally having minimal data content ( e.g. Start/Stop with no data or just list of peers ) , stateless in nature and immediate action on it.  SendReplayUpdates on the other hand for session(s) in Established/Up state might mean 
i) lots of data needing to be transmitted to test ports at run time when test is already running  ; 
ii) ambiguity as to whether the Updates can be sent at all or not ( e.g. BGP Opens being exchanged and session not yet up)  or even in Down state 
iii) conflict between ReplayUpdates part of SetConfig and those sent via trigger --> BGP implementations then need to start keeping multiple copies of different ReplayUpdates,
iv) Behaviour on session flap --> Do we send the routes once again ? [ Triggers are normally stateless and should not need to retain states once the trigger action is completed ; probably this can also impact GR triggers ] 
v) GetConfig will not return the _triggered_ ReplayUpdates since they are not part of config. 
ReplayUpdates as part of the initial config pre Start avoids these issues which is why I wanted to avoid post Start Control Update based trigger for this and instead attempt to have this as part of 
i) initial SetConfig -> use go library provided by OTG to convert proto binary to OTG structured format or create programatically. 
ii) if at all a ControlAction --> after SetConfig but before Start but only to make it as part of Config mainly for non golang users who cannot use the go library. ( after stability with this initial PR and at least one working implementation and some feedback from end-users ) 
Maybe we can have a discussion on what use-case/advantages you have in mind for scenario where the ReplayUpdates is not part of initial SetConfig but is replayed via trigger/Control State at a later point of time.

apratimmukherjee:
@jasdeep-hundal , @kingshukdev , @SuryyaKrJana , @SouravSinhaRoy , @alakendu , @Vibaswan : 
I have uploaded a new version of the model with following changes in anticipation of merging to models/main to be able to provide the initial implementation drop . These are the following changes:  
1) Removed all references to .proto definitions in the Descriptions. 
2) Modified MP_REACH and MP_UNREACH to have name based specification of AFI/SAFI as well as stricter type check for contents based on chosen type e.g. ipv4_unicast type would automatically map to AFI 1 SAFI 1 and value contents of that would contain only attributes specific to IPv4 Unicast. 
This was not the case earlier where the attributes were an union of attributes for all supported AFI/SAFIs. This would help to easily add support for more AFI/SAFIs without creating an union of unnecessary attributes e.g. when adding Segment Routing AFI/SAFI a new type will be added which in turn will have its own data contents containing extra attributes needed only for Segment Routing prefix such as Distinguisher and Color which is applicable only for Segment Routing and not for IPv4/Ipv6 Unicast . 
3) Removed the Tunnel TLV and all Segment Routing related attributes . Will share a new PR on top of this once this  PR is merged to main. This is to ensure that we can have a working implementation against it to detect issues if any before adding it to main . This is to minimize chances of need for backward compatibility breaking changes. 
4) Updated the Description to align with current decided workflow where the .protos and library will not be made public.
5) Update the code snippets in the Description to be in sync with the latest proposed version of the model.

@jasdeep-hundal : Request you to kindly review this and approve it if it looks okay. Please ping me via mail/slack or comments in the PR itself if you wish to clear up any doubts before we merge this to main.
Requesting everybody else in review list to also take a look and provide comments in the PR / reach out to me as needed so that we can converge on this quickly , to be able to provide an initial release based on this model as soon as possible.

jasdeep-hundal:
LGTM

PR #344, opened 2023-11-08
Allow BGP to be configured on a non-default TCP port.

SuryyaKrJana:
https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/bgp-listen-port/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config

This PR is for a sub-part of [RT-1.9: BGP Transport Parameters test](https://github.com/openconfig/featureprofiles/blob/bfe41d0d3231c3404ad993ae54ba159c46a703b2/feature/bgp/transport_tests/README.md#rt-19-bgp-transport-parameters-test).

Exposing BGP listen port number and neighbor port for user to change to non-default port number, say 1800 or 2000, as per need.

Configuration snippet:

bgpIpv4Interface := device.Bgp().Ipv4Interfaces().
                            Add().
                            SetIpv4Name(ipv4Name)
bgpPeer := bgpIpv4Interface.Peers().
                    Add().
                    SetAsNumber(uint32(ixiaAs)).
                    SetAsType(gosnappi.BgpV4PeerAsType.IBGP).
                    SetPeerAddress(peerRemoteIp).
                    SetName(peerName)

**bgpPeer.Advanced().SetListenPort(1800)**
**bgpPeer.Advanced().SetNeighborPort(2000)**

### DIFF
--- a/device/bgp/bgpadvanced.yaml
+++ b/device/bgp/bgpadvanced.yaml
@@ -54,7 +54,7 @@ components:
           x-field-uid: 6
         listen_port:
           description: |-
-            Set listen transport port number for the local BGP peer.
+            The TCP port number on which to accept BGP connections from the remote peer.
           type: integer
           format: uint32
           default: 179
--- a/device/bgp/bgpadvanced.yaml
+++ b/device/bgp/bgpadvanced.yaml
@@ -47,14 +47,12 @@ components:
           x-field-uid: 5
         passive_mode:
           description: |-
-            If enabled then the BGP Finite State Machine will passively wait 
-            for the remote BGP peer to establish the BGP TCP connection. 
-            If the passive mode is true, BGP peer will wait for the remote BGP peer 
-            to issue an Open request before an Open message is sent.
+            If set to true, the local BGP peer will wait for the remote peer to initiate the BGP session
+            by establishing the TCP connection, rather than initiating sessions from the local peer.
           type: boolean
           default: false
           x-field-uid: 6
-        local_listen_port:
+        listen_port:
           description: |-
             Set listen transport port number for the local BGP peer.
           type: integer
@@ -62,3 +60,12 @@ components:
           default: 179
           maximum: 65535
           x-field-uid: 7
+        neighbor_port:
+          description: |-
+            Destination TCP port number of the BGP peer when initiating a
+            session from the local BGP peer.
+          type: integer
+          format: uint32
+          default: 179
+          maximum: 65535
+          x-field-uid: 8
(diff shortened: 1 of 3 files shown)


device/bgp/bgpadvanced.yaml
```diff
@@ -45,3 +45,20 @@ components:
...
+        passive_mode:
+          description: |-
+            If enabled then the BGP Finite State Machine will passively wait 
+            for the remote BGP peer to establish the BGP TCP connection. 
+            If the passive mode is true, BGP peer will wait for the remote BGP peer 
+            to issue an Open request before an Open message is sent.
+          type: boolean
+          default: false
+          x-field-uid: 6
+        local_listen_port:
```
apratimmukherjee, line 57: I think the attribute defined here is expected to do two jobs : 
1) Define the listening port used to accept connections from peer .
2) Define the port used to connect to remote peer when passive_mode is false.
If this is the expectation, the name should probably be changed to listen_port to remove the assumption that it will be used only for local_listen_port and not when connecting to remote peer.
Otherwise , we need two attributes , one for local_listen_port and one for neighbor_port .

device/bgp/bgpadvanced.yaml
```diff
@@ -45,3 +45,20 @@ components:
...
+          description: |-
+            Set listen transport port number for the local BGP peer.
```
apratimmukherjee, line 59: Assuming this new attribute is used for both connecting and accepting BGP connections , the description may need to be updated to something like : 
Set the TCP port number to be used for both accepting BGP connections from remote peer as well as to initiate connections to remote peer.  [ Else for two separate attributes we need to break the description accordingly as well ]. 
Note sure if we need to specifically mention that if an implementation has limitations e.g. all sessions initiated from same test port needing to have same port number , set_config  should return error specifying the same.
Maybe if not in the model, this can be specified in PR notes before sending to wider team.

apratimmukherjee (changes requested):

device/bgp/bgpadvanced.yaml
apratimmukherjee: Also needs a linked Issue as discussed and code snippet after changes if any can be put inside code format in description.

jasdeep-hundal (approved):
No need to document the limitation that all peers on the same physical port need to be listening on the same TCP ports, that should be in the notes for Ixia-C/it's fair to just return an error there.

device/bgp/bgpadvanced.yaml
```diff
@@ -52,4 +52,20 @@ components:
           type: boolean
           default: false
           x-field-uid: 6
-        
+        listen_port:
+          description: |-
+            Set listen transport port number for the local BGP peer.
```
apratimmukherjee, line 57: The TCP port number on which to accept BGP connections from the remote peer.

apratimmukherjee (approved):
Minor desc. comment.

PR #347, opened 2023-11-23
Support for SNMPv2C Packet Header (Data Plane Only)

biplamal:
Redocly View:
[ReDoc Interactive Demo (redocly.github.io)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-snmpv2/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config)
( New object at : set_config/flows/packet/snmpv2c )

[SNMPv2-RFC3416](https://www.ietf.org/rfc/rfc3416.txt?number=3416)
[SNMPv2C-RFC1901](https://datatracker.ietf.org/doc/html/rfc1901)
[SNMP-TCP-Guide](http://www.tcpipguide.com/free/t_SNMPVersion2SNMPv2MessageFormats-3.htm)
[SNMP-OID-Encoding-Example](https://www.ranecommercial.com/legacy/note161.html)

Requirement:
The user able to define SNMPv2C data plane packet headers. Purpose is to flood DUT with SNMPv2C request and responses. The user should be able to configure different fields in PDU with ability to specify Increment/Decrement operations on certain applicable fields  in the specified PDU.

Example usage:

```
f1 := config.Flows().Items()[0]
f1.SetName("f1:p1->p2").
TxRx().Port().SetTxName(p1.Name()).SetRxName(p2.Name())

f1.Rate().SetPps(10)

f1.Packet().Add().Ethernet()
f1.Packet().Add().Ipv4()
f1.Packet().Add().Tcp()

// sample SNMPv2C get-request
SNMPv2C_1 := f1.Packet().Add().Snmpv2c()
SNMPv2C_1.SetCommunity("abc")
SNMPv2CGetReq := SNMPv2C_1.Data().GetRequest()
SNMPv2CGetReq .RequestId().SetValues([]uint32{1, 2})
varBind1 := SNMPv2CGetReq .VariableBindings().Add()
varBind1.SetObjectIdentifier([]uint64{1, 0, 8802, 1, 1, 1, 1, 1, 1, 0})

varBind2 := SNMPv2CGetReq .VariableBindings().Add()
varBind2.SetObjectIdentifier([]uint64{1, 0, 200, 1, 1, 1, 1, 1, 1, 0})

// sample SNMPv2C response
SNMPv2C_2 := f1.Packet().Add().Snmpv2c()
SNMPv2C_2.SetCommunity("abc")
SNMPv2CResp := SNMPv2C_2 .Data().Response()
SNMPv2CResp .RequestId().SetValues([]uint32{1, 2})
varBind1 := SNMPv2CResp.VariableBindings().Add()
varBind1.SetObjectIdentifier([]uint64{1, 0, 8802, 1, 1, 1, 1, 1, 1, 0})
varBind1.Value().Timeticks().Increment().SetStart(20).SetStep(2)

varBind2 := SNMPv2CResp.VariableBindings().Add()
varBind2.SetObjectIdentifier([]uint64{1, 0, 200, 1, 1, 1, 1, 1, 1, 0})
varBind2.Value().Counter().Decrement().SetStart(100).SetStep(1)
```


flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+      description: A Sequence of two fields, an object_identifier and the value for/from that object_identifier.
+      type: object
+      properties:
+        object_identifier:
+          description: >-
+              object_identifier that points to a particular parameter in the SNMP agent.
+          type: array
+          items:
+            type: integer
+            format: int64
```
apratimmukherjee, line 136: Is this 64 bits in PDU ?

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+              x-field-uid: 3
+            arbitrary:
+              x-field-uid: 4
+            big_counter:
+              x-field-uid: 5
+            unsigned_integer:
+              x-field-uid: 6
+        ip_address:
+          description: >-
+            octet string
```
apratimmukherjee, line 164: I understand it can contain both ipv4 and ipv6 address ? Maybe we can have better strict data types by asking for v4/v6 type and then specific address ?

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+          type: string
+          x-field-uid: 2
+        counter:
+          type: integer
+          format: int64
```
apratimmukherjee, line 169: Should this fields not be of type uint64 ? Why int64 ? and why min / max is specified . Maybe they should have a default value of 0  else should they not be required fields ?

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+          minimum: 0
+          maximum: 4294967295
+          x-field-uid: 3
+        timeticks:
+          type: integer
+          format: int64
```
apratimmukherjee, line 175: Similar comment as for other int64 types

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+          minimum: 0
+          maximum: 4294967295
+          x-field-uid: 4
+        arbitrary:
+          description: >-
+            octet string
```
apratimmukherjee, line 181: octet string is not clear ... do you mean ascii text ?

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
...
+          type: string
+          x-field-uid: 5
+        big_counter:
+          type: integer
+          format: int64
```
apratimmukherjee, line 186: Similar comment as other int64 fields. Does this have some max hex value specified somewhere. If yes, maybe explain that. But still should probably be uint64 ?

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,201 @@
+components:
+  schemas:
+    Flow.Snmpv2:
+      description: SNMPv2 packet header
```
apratimmukherjee, line 4: If we do v3 later , would we need to repeat everything there ? Or can v2 definition be used with v3 as a superset on top of that ?

apratimmukherjee (commented):
Initial review.

flow/packet-headers/snmpv2.yaml
biplamal: Added

flow/packet-headers/snmpv2.yaml
biplamal: Done

flow/packet-headers/snmpv2.yaml
biplamal: Done

flow/packet-headers/snmpv2.yaml
biplamal: Done

flow/packet-headers/snmpv2.yaml
biplamal: Discussed

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,288 @@
...
+        value: 
+          $ref: '#/components/schemas/Flow.Snmpv2.VariableBindingValue'
+          x-field-uid: 2
+    Flow.Snmpv2.VariableBindingValue: 
+      description: >-
+          The value for the object_identifier as per RFC2578.
+      type: object
+      properties:
+        choice:
+          type: string
```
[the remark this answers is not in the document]
biplamal: Done

rudranil-das (approved):
LGTM

flow/packet-headers/snmpv2.yaml
```diff
@@ -0,0 +1,308 @@
...
+            signed: true
+            features: [count]
+          x-field-uid: 1
+        non_repeaters:
+          x-field-pattern:
+            description: >-
+                One variable binding in the Response-PDU is requested for the first non_repeaters variable bindings in the GetBulkRequest.
+            format: integer
+            length: 32
+            default: 0
```
PrasenjitAdhikary, line 156: Any specific reason to skip the counter (increment/decrement) pattern for this field?

apratimmukherjee: @PrasenjitAdhikary : Having this as variable basically means that the different packets in the flow would normally have different sizes due to different number of OIDs .. where no. of OIDs ( variable bindings ) >= value in this field. I think it is not possible to do that inside a single flow and will be difficult to actually use to create a flow with multiple valid packets with this field having different values ( similar to say trying to do this with length field in IP packets ) ... Just ability to put any value in this field should also cover negative use-cases .

biplamal: Yes true, that is the logic behind not having counter on this field, @PrasenjitAdhikary

PrasenjitAdhikary (approved):

apratimmukherjee (approved):

ashutshkumr:
Can't we accept object identifier as string ?

The reason being, if we were to accept array of uint64 as input, it would complicate how multiple values will be provided.

e.g. `"1.0.8802.1.1.1.1.1.1.0"` vs `[]uint64{1, 0, 8802, 1, 1, 1, 1, 1, 1, 0}`

flow/packet-headers/snmpv2c.yaml
```diff
@@ -0,0 +1,366 @@
...
+              x-field-uid: 2
+        ipv4:
+          x-field-pattern:
+            description: >-
+              IPv4 address returned for the requested OID.
+            format: ipv4
+            default: 0.0.0.0
+            features: [count]
+          x-field-uid: 2
+        ipv6:
```
alakendu, line 314: According to [SNMPv2-RFC3416](https://www.ietf.org/rfc/rfc3416.txt?number=3416) supported `ApplicationSyntax` are

```
   ApplicationSyntax ::= CHOICE {
         ipAddress-value        IpAddress,
         counter-value          Counter32,
         timeticks-value        TimeTicks,
         arbitrary-value        Opaque,
         big-counter-value      Counter64,
         unsigned-integer-value Unsigned32 }
```
SNMP Data Type is supporting single IP type (`IpAddress`) with `SIZE (4)`
```
IpAddress ::= [APPLICATION 0] IMPLICIT OCTET STRING (SIZE (4))
```
So, I am suspecting RFC talked about for IPv4 address NOT IPv6 address. Because if this is IPv6 then what is the length of it.

Could you please double check it otherwise we have to remove IPv6.

flow/packet-headers/snmpv2c.yaml
```diff
@@ -0,0 +1,366 @@
...
+              x-field-uid: 6
+            timeticks_value:
+              x-field-uid: 7
+            arbitrary_value:
+              x-field-uid: 8
+            big_counter_value:
+              x-field-uid: 9
+            unsigned_integer_value:
+              x-field-uid: 10
+        integer_value:
```
alakendu, line 225: We are proving OTG data type (e,g, `<int32>`, <uint32>`, `<string>` etc) within all SNMP property. But those data types are not one-to-one mapping with ASN.1 type. say 
- `community` and `error_status` are `string` but those are different ASN.1 type (which are `0x04` and `0x02`)
- Same also true for `request_id` and `integer_value`, both are `int32` with different ASN.1 type (`0x02` and `0x41`)

Yes, I agree this is not a goal for OTG model perspective. So, I am thinking it will better to mentioned what is the ANS.1 type for each field within description. It will reduced ramp-up and anomality at the time to develop SNMP protocol.

ashutshkumr:
The `value` suffix in all the properties and choices inside `variable binding` seems redundant. Is this suffix suggested by RFC ? If not, I would very much propose to get rid of these.

biplamal:
> The `value` suffix in all the properties and choices inside `variable binding` seems redundant. Is this suffix suggested by RFC ? If not, I would very much propose to get rid of these.

Yes its suggested by rfc

PR #349, opened 2023-12-06
IP options in IPv4 header

indraniBh:
ReDocly link: 
[ReDoc Interactive Demo (redocly.github.io)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-ip-options/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config)

( New object at : set_config/flows/packet/ipv4/options )

[IPv4 RFC: RFC791](https://datatracker.ietf.org/doc/html/rfc791)
[Router Alert Option RFC: RFC2113](https://datatracker.ietf.org/doc/html/rfc2113)
[IPv4 Options parameters](https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml#ip-parameters-1)

Objective:  IPv4 options are optional extensions for the IPv4 header that can be utilised to provide additional information about the IPv4 datagram. It is encoded as a series of type, length and value attributes. The IP header length MUST be increased to accommodate the extra bytes needed to encode the IP options. 
Currently IP options supported are: 
1. router_alert Option allows devices to intercept packets not addressed to them directly as defined in RFC2113. 
2. custom option is provided for to be able to configure user defined IP options as needed.
IPv4 Options is an array option objects.
To support RSVP control packet, router alert option is required in IP header.

Sample IPv4 Options (Router Alert and Custom):

```
        srcIP := "10.10.0.1"
	dstIP := "10.10.0.2"

	// add ports
	p1 := config.Ports().Add().SetName("p1").SetLocation("eth1")
	p2 := config.Ports().Add().SetName("p2").SetLocation("eth2")

	f1 := config.Flows().Items()[0]
	f1.SetName("f1:p1->p2").
		TxRx().Port().SetTxName(p1.Name()).SetRxName(p2.Name())
	f1.Rate().SetPps(10)

	f1.Packet().Add().Ethernet()
	ip := f1.Packet().Add().Ipv4()

	ip.Src().SetChoice("value").SetValue(srcIP)
	ip.Dst().SetChoice("value").SetValue(dstIP)

	// Sample of router_alert option:
	ip.Options().Add().SetChoice("router_alert")

	// Sample of user defined custom TLV options (Stream ID)
	ipOptionCustom := ip.Options().Add().SetChoice("custom")
	ipOptionCustom.Custom().Type().CopiedFlag().SetChoice("value").SetValue(0)
	ipOptionCustom.Custom().Type().OptionClass().SetChoice("value").SetValue(0)
	ipOptionCustom.Custom().Type().OptionNumber().SetChoice("value").SetValue(8)

	ipOptionCustom.Custom().Length().SetChoice("value").SetValue(4)
	ipOptionCustom.Custom().SetValue("0088")
```

### DIFF
--- a/flow/packet-headers/ipv4.yaml
+++ b/flow/packet-headers/ipv4.yaml
@@ -151,7 +151,7 @@ components:
               x-field-uid: 2
         custom:
           $ref: '#/components/schemas/Flow.Ipv4Options.Custom'
-          x-field-uid: 3
+          x-field-uid: 2
     Flow.Ipv4Options.Custom:
       description: >-
         User defined IP options to be appended to the IPv4 header.
--- a/flow/packet-headers/ipv4.yaml
+++ b/flow/packet-headers/ipv4.yaml
@@ -122,6 +122,101 @@ components:
             default: 0.0.0.0
             features: [count, metric_tags]
           x-field-uid: 14
+        options:
+          $ref: '#/components/schemas/Flow.Ipv4.Options'
+          x-field-uid: 15
+    Flow.Ipv4.Options:
+      description: >-
+        IP options for Router Alert Options (RFC 2113) and custom TLV
+      type: object
+      properties:
+        choice:
+          type: string
+          default: router_alert
+          x-field-uid: 1
+          x-enum:
+            router_alert:
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        router_alert:
+          x-field-pattern:
+            description: >-
+              Router Alert Option can intercept packets not addressed to them directly
+            format: integer
+            length: 16
+            default: 0
+            features: [count, metric_tags]
+          x-field-uid: 2
+        custom:
+          $ref: '#/components/schemas/Flow.Ipv4.Custom'
+          x-field-uid: 3
+    Flow.Ipv4.Custom:
+      description: >-
+        IP options for custom TLV fields
+      type: object
+      properties:
+        type:
+          $ref: '#/components/schemas/Flow.Ipv4.Type'
+          x-field-uid: 1
+        length:
+          x-field-pattern:
+            description: >-
+              Length of the option field
+            format: integer
+            length: 8
+            default: 4
+            features: [count, metric_tags]
+          x-field-uid: 2
+        value:
+          x-field-pattern:
+            description: >-
+              Length of the option field
+            format: integer
+            length: 16
+            default: 0
+            features: [count, metric_tags]
+          x-field-uid: 3
+    Flow.Ipv4.Type:
+      description: >-
+        Type options for custom TLV field in Options
+      type: object
+      properties:
+        copied_flag:
+          x-field-pattern:
+            x-constants:
+              not_copied: 0
+              copied: 1
+            description: >-
+              This flag indicates this option is copied to all fragments on fragmentations
+            format: integer
+            length: 1
+            default: 0
+            features: [count, metric_tags]
+          x-field-uid: 1
+        option_class:
+          x-field-pattern:
+            x-constants:
+              control: 0
+              reserved: 1
+              debugging_measurement: 2
+              reserved: 3
+            description: >-
+              Option class
+            format: integer
+            length: 2
+            default: 0
+            features: [count, metric_tags]
+          x-field-uid: 2
+        option_number:
+          x-field-pattern:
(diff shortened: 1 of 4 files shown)


rudranil-das:
Please update example by actually adding item in `ip.Options()` for usage in each of the choices

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+            description: >-
+              Length of the option field
+            format: integer
+            length: 16
+            default: 0
+            features: [count]
+          x-field-uid: 3
+    Flow.Ipv4.Type:
+      description: >-
+        Type options for custom TLV field in Options
```
rudranil-das, line 181: I think the term `TLV` should be avoided here.

rudranil-das (approved):
rest LGTM

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+          items:
+            $ref: '#/components/schemas/Flow.Ipv4.Options'
+          x-field-uid: 15
+    Flow.Ipv4.Options:
+      description: >-
+        IPv4 options are optional extensions for the IPv4 header that can be utilised to provide additional information about the IPv4 datagram. 
+        It is encoded as a series of type, length and value attributes. 
+        The IP header length MUST be increased to accommodate the extra bytes needed to encode the IP options.
+        Currently IP options supported are:
+        1. router_alert Option allows devices to intercept packets not addressed to them directly as defined in RFC2113.
```
apratimmukherjee, line 137: Option -> option

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+        2. custom option is provided for to be able to configure user defined IP options as needed.
```
apratimmukherjee, line 138: is provided for to be able to configure-> is provided to configure

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+          x-field-uid: 1
+          x-enum:
+            router_alert:
+              x-field-uid: 1
+            custom:
+              x-field-uid: 2
+        custom:
+          $ref: '#/components/schemas/Flow.Ipv4.Custom'
+          x-field-uid: 3
+    Flow.Ipv4.Custom:
```
apratimmukherjee, line 153: Not sure if the schema is okay to be Flow.Ipv4.Custom  or better to make it  Flow.Ipv4Options.Custom

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+      description: >-
+        User defined IP options to be appended to the IPv4 header.
+      type: object
+      properties:
+        type:
+          $ref: '#/components/schemas/Flow.Ipv4.Type'
```
apratimmukherjee, line 159: Not sure if the schema is okay to be Flow.Ipv4.Type or better to make it  Flow.Ipv4Options.Type

flow/packet-headers/ipv4.yaml
```diff
@@ -122,6 +122,100 @@ components:
...
+            format: integer
+            length: 8
+            default: 4
+            features: [count]
+          x-field-uid: 2
+        value:
+          x-field-pattern:
+            description: >-
+              Length of the option field
+            format: integer
```
apratimmukherjee, line 174: Since this is a custom option, the value field should probably be a string restricted to take input of a hex string to truly allow custom options to be added. We will either need to throw error/warning  in implementation if hex bytes length do not match with length and auto update both options length and IP header length when creating packet OR just create the packet as provided and specify explicitly that for custom option , IP header length needs to be set explicitly by user to create valid packet .  Probably 1st is more user-friendly but second allows for creation of back IP packets with incorrect length etc. with IP options + IP options with bad length etc. 
Probably providing auto option for length with value override might handle both .

apratimmukherjee:
Please update the example to reflect the changes for adding custom option. Also need to actually add option to packet template I think as @rudranil-das had mentioned to make it a valid snippet.

indraniBh:
Updated code snippet. @rudranil-das @apratimmukherjee

PR #370, opened 2024-03-26
Support for User Defined Metrics

Vibaswan:
**Redocly View**
[ReDoc Interactive Demo (redocly.github.io)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-uds/artifacts/openapi.yaml&nocors#tag/Monitor)

New Objects at:
- `set_config/filter_metrics`
- `get_metrics/user_defined`

**Requirement**
Offer a means for users to configure and retrieve metrics based on custom conditions established by the user on packets generated by the test tool. These filters can be specified either on a per-port or per-flow basis.

**Usage**
```go
// configure filter metrics
f1 := config.FilterMetrics().Add()
f1.SetName("filter_ip").Association().SetPortNames([]string{"p1", "p2"})
f1.ReceiveFilters().Add().Ipv4().Dst().SetValue("01010101").SetMask("00000000")
f1.ReceiveFilters().Add().Custom().SetValue("01010101").SetMask("00000000").SetOffset(18)

// rest of the script

mr := gosnappi.NewMetricsRequest()
mr.UserDefined().SetPortNames([]string{"p1", "p2"})
api.GetMetrics(mr)

/*

Result structure
------------------------------------
Response:
	- name: p1
	  frames_tx: 1000
	  frames_rx: 0
	  filtered_metrics:
	    -  filter_name: "filter_ip"
	       filter_value: 0
	- name: p2
	  frames_tx: 0
	  frames_rx: 1000
	  filtered_metrics:
	    -  filter_name: "filter_ip"
	       filter_value: 500

Result of the metrics
-------------------------------------
Name	frames_tx	    frames_rx	filter_ip
p1	        1000	    0			0
p2		0		   1000		500

*/
```

### DIFF
--- a/filter_metrics/filter_metrics.yaml
+++ b/filter_metrics/filter_metrics.yaml
@@ -43,7 +43,6 @@ components:
           description: |-
             The type of association.
           type: string
-          default: port_names
           x-field-uid: 1
           x-enum:
             port_names:
--- a/filter_metrics/filter_metrics.yaml
+++ b/filter_metrics/filter_metrics.yaml
@@ -1,0 +1,72 @@
+components:
+  schemas:
+    FilterMetric:
+      x-status:
+        status: under_review
+        information: There may be changes in filter_metric configuration
+      description: >-
+        Configuration for FilterMetrics settings.
+      type: object
+      required: [name]
+      properties:
+        association:
+          description: >-
+            Association.
+          $ref: '#/components/schemas/FilterMetric.Association'
+          minItems: 1
+          maxItems: 32
+          x-field-uid: 1
+        receive_filters:
+          description: >-
+            A list of filters to apply to the packet receiving ports.
+            If no filters are specified then all packets will be counted.
+            A filter_metric can have multiple filters. The number of filters supported
+            is determined by the implementation.
+
+            When multiple filters are specified then implementation must &&
+            (and) all the filters.
+          type: array
+          items:
+            $ref: '../capture/capture.yaml#/components/schemas/Capture.Filter'
+          minItems: 1
+          maxItems: 32
+          x-field-uid: 2
+        name:
+          x-include: ../common/common.yaml#/components/schemas/Named.Object/properties/name
+          x-field-uid: 3
+
+    FilterMetric.Association:
+      description: >-
+        Configuration for filter_metric Association.
+        Based on the association, filter_metric will be applied to count packets on per port or per flow.
+      type: object
+      properties:
+        choice:
+          description: |-
+            The type of association.
+          type: string
+          default: port_names
+          x-field-uid: 1
+          x-enum:
+            port_names:
+              x-field-uid: 1
+            flow_names:
+              x-field-uid: 2
+        port_names:
+          description: >-
+            List of port names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+          x-field-uid: 2
+        flow_names:
+          description: >-
+            List of flow names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - '/components/schemas/Flow/properties/name'
+          x-field-uid: 3   
--- a/filter_metrics/filter_metrics.yaml
+++ b/filter_metrics/filter_metrics.yaml
@@ -25,7 +25,7 @@ components:
             (and) all the filters.
           type: array
           items:
-            $ref: '../capture/capture.yaml#/components/schemas/Capture.Filter'
+            $ref: '#/components/schemas/Filter.Packet'
           minItems: 1
           maxItems: 32
           x-field-uid: 2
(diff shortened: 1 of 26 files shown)


PrasenjitAdhikary:
In the Usage description, we need to add a flow example. 
In the example for the result, aggregated tx / rx are not retrieved through filter stats and should not be part of the UDS response.

Vibaswan:
> In the Usage description, we need to add a flow example. In the example for the result, aggregated tx / rx are not retrieved through filter stats and should not be part of the UDS response.

Corrected the usage description according to suggestion

result/uds.yaml
```diff
@@ -0,0 +1,101 @@
...
+        The user defined metrics result request to the traffic generator
+      type: object
+      properties:
+        choice:
+          description: |-
+            The type
+          type: string
+          x-field-uid: 1
+          x-enum:
+            port_names:
```
lyle-thompson-keysight, line 14: why not use PortMetrics.Association here?

lyle-thompson-keysight:
In the example in the overview, the masks should not be 0.

filter_metrics/filter_metrics.yaml
```diff
@@ -0,0 +1,222 @@
...
+      properties:
+        value:
+          type: string
+          format: hex
+          default: "00"
+          x-field-uid: 1
+        mask:
+          type: string
+          format: hex
+          default: "00"
```
lyle-thompson-keysight, line 221: default mask should be 0xff. But there could be problems with field width (i.e. field width varies with field). Perhaps we need a special variable-width hex value, e.g. 0xf..., or !0. If that won't work, we should not give mask a default value, because 00 is (almost?) never right.

result/uds.yaml
```diff
@@ -0,0 +1,101 @@
...
+          description: |-
+            Names of flows for which to return results.
+            An empty list will return all flow row results.
+          type: array
+          items:
+            type: string
+          x-constraint:
+            - '/components/schemas/Flow/properties/name'
+          x-field-uid: 3
+        column_names:
```
lyle-thompson-keysight, line 38: I don't quite see why we need column_names here. Wouldn't we always return a column for each stat that we requested? And user can reduce by not requesting specific UDS values?

result/uds.yaml
```diff
@@ -0,0 +1,101 @@
+components:
+  schemas:
+    User.Defined.Metrics.Request:
```
lyle-thompson-keysight, line 3: The use of the term "User Defined Stats" seems  a bit odd, given that you've defined a large number of built-in stats, including an option of "custom". "custom" and "user-defined" are essentially synonymous, so perhaps we should look for a better name.

filter_metrics/filter_metrics.yaml
Vibaswan: Removed default value for mask

Vibaswan:
> In the Usage example, the masks should not be 0.

Corrected the usage description

result/uds.yaml
Vibaswan: Introduced it so that it is at par with all the other stat requests , as all of them have column names , but it makes sense to not have it for UDS as the choice is right there with the users.
Removed column names

result/uds.yaml
Vibaswan: we removed column names and now we are reusing association

result/uds.yaml
Vibaswan: Currently we are using this 
```go 
gosnappi.NewMetricsRequest().UserDefined()
```
Here are some alternatives I was thinking

```go 
gosnappi.NewMetricsRequest().Customized()
```
```go 
gosnappi.NewMetricsRequest().Tailored()
```
```go 
gosnappi.NewMetricsRequest().Targeted()
```

Let me know if you have some suggestion or you would like one of the above.

PrasenjitAdhikary (commented):
Bin size for UHD400 is 1000, unless there is any further update from UHD team.

anjan-keysight:
Example given in usage is wrong; for every receive bin we can have only one header field (other than src and dst mac). So configuring both IPv4->dst and custom fields for rb1 in usage is wrong. You can give two separate examples.

anjan-keysight:
How do we specify the correct header field for receive bins if we have say multiple headers of same type (say inner and outer IPv4 and I want to set receive bins only for inner IPv4 dst field).

anjan-keysight:
In the get metrics, we specify the list of ports and also the list of receive bins. What if some bins does not apply to some ports. Will we silently ignore those bins for that port, or return error. Not all bins will apply to all specified ports.

Secondly, the include flag is redundant I think. If we specify the bin name, means we want to include it.

Vibaswan:
> Bin size for UHD400 is 1000, unless there is any further update from UHD team.

According to a communication with Lyle we would first Target 6 and then later see if we can increase the number

Vibaswan:
> Example given in usage is wrong; for every receive bin we can have only one header field (other than src and dst mac). So configuring both IPv4->dst and custom fields for rb1 in usage is wrong. You can give two separate examples.

Corrected the example in the description

Vibaswan:
> How do we specify the correct header field for receive bins if we have say multiple headers of same type (say inner and outer IPv4 and I want to set receive bins only for inner IPv4 dst field).

Added a example for Ip in Ip in the description

Vibaswan:
> In the get metrics, we specify the list of ports and also the list of receive bins. What if some bins does not apply to some ports. Will we silently ignore those bins for that port, or return error. Not all bins will apply to all specified ports.
> 
> Secondly, the include flag is redundant I think. If we specify the bin name, means we want to include it.

removed redundant fields from metrics request , 
I feel it it does not find a match implementation should ignore, I am not aware of any metrics response which returns error if it does not find a perfect match @PrasenjitAdhikary would please let us know your views on this

laurchoy:
> > In the get metrics, we specify the list of ports and also the list of receive bins. What if some bins does not apply to some ports. Will we silently ignore those bins for that port, or return error. Not all bins will apply to all specified ports.
> > Secondly, the include flag is redundant I think. If we specify the bin name, means we want to include it.
> 
> removed redundant fields from metrics request , I feel if it does not find a match, implementation should ignore, I am not aware of any metrics response which returns error if it does not find a perfect match @PrasenjitAdhikary would please let us know your views on this


If the port doesn't find a match, there will be a default bin that it will drop into for all packets not matching any criteria. This way you can match back where the packet went. Also, counts should not be duplicated across bins. This is the user responsibility to make sure of that. The total number of packets across all bins should always equal to number of packets received.

PrasenjitAdhikary:
> > > In the get metrics, we specify the list of ports and also the list of receive bins. What if some bins does not apply to some ports. Will we silently ignore those bins for that port, or return error. Not all bins will apply to all specified ports.
> > > Secondly, the include flag is redundant I think. If we specify the bin name, means we want to include it.
> > 
> > 
> > removed redundant fields from metrics request , I feel if it does not find a match, implementation should ignore, I am not aware of any metrics response which returns error if it does not find a perfect match @PrasenjitAdhikary would please let us know your views on this
> 
> If the port doesn't find a match, there will be a default bin that it will drop into for all packets not matching any criteria. This way you can match back where the packet went. Also, counts should not be duplicated across bins. This is the user responsibility to make sure of that. The total number of packets across all bins should always equal to number of packets received.

This behavior is not available in otg-hw or ixia-c (usstream). All UDS are independent, and there is no concept of a default bin, for un-matched packets. We cannot change this behavior in hardware/ixia-c. 
So, to support minimum features across all platforms, UHD400 needs to support this behavior (I believe the same is already happening for UHD-100). Why this cannot be supported in UHD400? 

We can add additional behavior specific to UHD only, if there is UHD UHD-specific requirement from the customer, in future sprints. But to support the UDS feature across all platforms uniformly, within the committed timeline, we should need this basic support, with independent bins.

laurchoy:
>UHD400 needs to support this behavior (I believe the same is already happening for UHD-100). Why this cannot be >supported in UHD400?
This has been discussed before. The decision was to have multiple pattern matches rather than independent matches. Jasdeep agreed to that as well.

PrasenjitAdhikary:
> > UHD400 needs to support this behavior (I believe the same is already happening for UHD-100). Why this cannot be >supported in UHD400?
> > This has been discussed before. The decision was to have multiple pattern matches rather than independent matches. Jasdeep agreed to that as well.

At that point, it was planned to have a different workflow for UHD (flow-level) & OTG-HW (port-level). Now we agreed to have a common minimum that all platforms will support.

laurchoy:
> At that point, it was planned to have a different workflow for UHD (flow-level) & OTG-HW (port-level). Now we agreed to have a common minimum that all platforms will support.

So let me explain the reasoning we took for having default bins and multiple matches rather than independent matches. Let's assume we have 100 packets receive. independent match A claim it received 50 and independent match B claimed it received 60.  This is a single test case where we receive packets from the DUT which can drop or replicate packets. From the received statistics, you cannot conclude that 50 packets that matches A matched B and the total number of packets that didn't match A or B just by looking at the receive packet count.  This is not the case for the decision that we had put in place to enforce this and also to help debug issues.

PR #391, opened 2024-09-27
Expose TCP MSS in BGP Peer

SuryyaKrJana:


apratimmukherjee:
This requirement is addressed for now by supporting MTU increase in ethernet level and modification of FP test accordingly. Closing .

PR #396, opened 2024-11-19
Correcting protocol field in GRE packet header by making it auto by default

### LINKED ISSUE #395: Correct GRE packet header protocol field to be auto by default
Its not possible to send a GRE packet without any protocol information if that protocol field is not specified in the packet header configuration , hence it needs to be auto by default

Vibaswan:
Fixes #395

### DIFF
--- a/flow/packet-headers/gre.yaml
+++ b/flow/packet-headers/gre.yaml
@@ -42,7 +42,7 @@ components:
             format: integer
             length: 16
             default: 2048
-            features: [auto, count, metric_tags]
+            features: [count, metric_tags, auto]
           x-field-uid: 4
         checksum:
           x-field-pattern:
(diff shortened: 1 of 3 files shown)


flow/packet-headers/gre.yaml
```diff
@@ -42,7 +42,7 @@ components:
             format: integer
             length: 16
             default: 2048
```
apratimmukherjee, line 44: Wondering if we change the default to auto. Assuming in a test today it is not set only if the below protocol matches type 2048 which will continue to work with new default but helps avoid user having to find the value for other protocol types.

apratimmukherjee (approved):
Suggested default change.

flow/packet-headers/gre.yaml
apratimmukherjee: Final generated yaml has auto by default.

PR #404, opened 2025-02-05
changed max vlan priority to 7

SuryyaKrJana:
Issue: [Athena-G](https://jira.it.keysight.com/browse/ATHENAG)[ATHENAG-2836](https://jira.it.keysight.com/browse/ATHENAG-2836)
Increased max value of vlan priority for device/ethernet/vlan from 3 to 7 to allow all 3 bits to be set.

### DIFF
--- a/device/vlan.yaml
+++ b/device/vlan.yaml
@@ -27,7 +27,7 @@ components:
             Priority code point
           type: integer
           format: uint32
-          maximum: 8
+          maximum: 7
           default: 0
           x-field-uid: 2
         id:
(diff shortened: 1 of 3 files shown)


device/vlan.yaml
```diff
@@ -27,7 +27,7 @@ components:
             Priority code point
           type: integer
           format: uint32
-          maximum: 3
+          maximum: 8
```
apratimmukherjee, line 30: maximum should be 7 for bits , 8 seems wrong.

apratimmukherjee (approved):

PR #414, opened 2025-03-12
One metric name change in MACsec model

sasubrata:
Rename metric name from bad_pkts_rx to in_pkts_bad in MACsec model


apratimmukherjee (approved):
LGTM

Issue #416, opened 2025-03-24
Add egress only tracking


indraniBh:
Add egress only tracking feature in configuration and corresponding metrics in Monitor.
This will be helpful for egress tracking per port flows in case of such protocols which has encrypted payload, e.g, macsec.

PR #417, opened 2025-03-24
Add support for egress only tracking

indraniBh:
**Issue:** [#416](https://github.com/open-traffic-generator/models/issues/416)

**Redocly View:**
[ReDoc Interactive Demo (redocly.github.io)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-egress-only/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config)

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -64,6 +64,13 @@ components:
           items:
             $ref: '../lldp/lldp.yaml#/components/schemas/Lldp'
           x-field-uid: 9
+        egress_only_tracking:
+          description: |-
+            IB: Egress only tracking.
+          type: array
+          items:
+            $ref: '../egress/egress-only-tracking.yaml#/components/schemas/EgressOnlyTacking'
+          x-field-uid: 10
     Config.Options:
       description: |-
         Global configuration options.
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -66,7 +66,7 @@ components:
           x-field-uid: 9
         egress_only_tracking:
           description: |-
-            IB: Egress only tracking.
+            The list of structures containing Egress offset and it's length to be placed in the packet. 
           type: array
           items:
             $ref: '../egress/egress-only-tracking.yaml#/components/schemas/EgressOnlyTacking'
--- a/egress/egress-only-tracking.yaml
+++ b/egress/egress-only-tracking.yaml
@@ -1,0 +1,6 @@
+components:
+  schemas:
+    EgressOnlyTacking:
+      description: |-
+        IB: Add description
+      type: object
--- a/egress/egress-only-tracking.yaml
+++ b/egress/egress-only-tracking.yaml
@@ -21,6 +21,11 @@ components:
           items: 
             $ref: '#/components/schemas/EgressOnlyTacking.MetricTag'
           x-field-uid: 2
+        filter:
+          description: |-
+            IB: Description of filter
+          $ref: '#/components/schemas/EgressOnlyTacking.Filter'
+          x-field-uid: 3
     EgressOnlyTacking.MetricTag:
       description: |-
         Metric Tag can be used to enable tracking portion of or all bits
@@ -53,3 +58,17 @@ components:
           default: 1
           minimum: 1
           x-field-uid: 3
+    EgressOnlyTacking.Filter:
+      type: object
+      properties:
+        choice:
+          description: >-
+            IB: Add description
+          type: string
+          x-field-uid: 1
+          default: none
+          x-enum:
+            none:
+              x-field-uid: 1
+            macsec:
+              x-field-uid: 2
--- a/egress/egress-only-tracking.yaml
+++ b/egress/egress-only-tracking.yaml
@@ -4,3 +4,52 @@ components:
       description: |-
         IB: Add description
       type: object
+      required: [port_name]
+      properties:
+        port_name:
+          description: Name of the Rx port where egress only tracking enabled.
+          type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+          x-field-uid: 1
+        metric_tags:
+          description: |-
+            One or more metric tags can be used to enable tracking portion of or all bits
+            in a corresponding header field for metrics per each applicable value.
+            These would appear as tagged metrics in corresponding flow metrics.
+          type: array
+          items: 
+            $ref: '#/components/schemas/EgressOnlyTacking.MetricTag'
+          x-field-uid: 2
+    EgressOnlyTacking.MetricTag:
+      description: |-
+        Metric Tag can be used to enable tracking portion of or all bits
+        in a corresponding header field for metrics per each applicable value.
+        These would appear as tagged metrics in corresponding flow metrics.
+      type: object
+      required:
+      - name
+      properties:
+        name:
+          description: |-
+            Name used to identify the metrics associated with the values applicable
+            for configured offset and length inside corresponding header field
+          type: string
+          pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
+          x-field-uid: 1
+        offset:
+          description: |-
+            Offset in bits relative to start of corresponding header field
+          type: integer
+          format: uint32
+          default: 0
+          x-field-uid: 2
+        length:
+          description: |-
+            Number of bits to track for metrics starting from configured offset
+            of corresponding header field
+          type: integer
+          format: uint32
+          default: 1
+          minimum: 1
+          x-field-uid: 3
--- a/egress/egress-only-tracking.yaml
+++ b/egress/egress-only-tracking.yaml
@@ -2,7 +2,7 @@ components:
   schemas:
     EgressOnlyTacking:
       description: |-
-        IB: Add description
+        The list of structures containing Egress offset and it's length to be placed in the packet.
       type: object
       required: [port_name]
       properties:
@@ -16,14 +16,14 @@ components:
           description: |-
             One or more metric tags can be used to enable tracking portion of or all bits
             in a corresponding header field for metrics per each applicable value.
-            These would appear as tagged metrics in corresponding flow metrics.
+            These would appear as tagged metrics in corresponding egress_only_tracking metrics.
           type: array
           items: 
             $ref: '#/components/schemas/EgressOnlyTacking.MetricTag'
           x-field-uid: 2
         filter:
           description: |-
-            IB: Description of filter
+            Offset in bits where the tracking value will be placed in the packet.
           $ref: '#/components/schemas/EgressOnlyTacking.Filter'
           x-field-uid: 3
     EgressOnlyTacking.MetricTag:
@@ -63,7 +63,7 @@ components:
       properties:
         choice:
           description: >-
-            IB: Add description
+            Offset in bits where the tracking value will be placed in the packet.
           type: string
           x-field-uid: 1
           default: none
--- a/result/egressonlytracking.yaml
+++ b/result/egressonlytracking.yaml
@@ -1,0 +1,471 @@
+components:
+  schemas:
+    EgressOnlyTracking.Metrics.Request:
+      description: >-
+        The container for a egress only tracking metric request.
+      type: object
+      properties:
+        port_names:
+          description: |-
+            Egress only tracking metrics will be retrieved for these port names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - "/components/schemas/EgressOnlyTacking/properties/port_name"
+          x-field-uid: 1
+        metric_names:
+          description: >-
+            The list of metric names that the returned result set will contain.
+            If the list is empty then all metrics will be returned.
+          type: array
+          items:
+            type: string
+            x-enum:
+              transmit:
+                x-field-uid: 1
+              frames_tx:
+                x-field-uid: 2
+              frames_rx:
+                x-field-uid: 3
+              bytes_tx:
+                x-field-uid: 4
+              bytes_rx:
+                x-field-uid: 5
+              frames_tx_rate:
+                x-field-uid: 6
+              frames_rx_rate:
+                x-field-uid: 7
+              tx_l1_rate_bps:
+                x-field-uid: 8
+              rx_l1_rate_bps:
+                x-field-uid: 9
+              tx_rate_bytes:
+                x-field-uid: 10
+              rx_rate_bytes:
+                x-field-uid: 11
+              tx_rate_bps:
+                x-field-uid: 12
+              rx_rate_bps:
+                x-field-uid: 13
+              tx_rate_kbps:
+                x-field-uid: 14
+              rx_rate_kbps:
+                x-field-uid: 15
+              tx_rate_mbps:
+                x-field-uid: 16
+              rx_rate_mbps:
+                x-field-uid: 17
+          x-field-uid: 2
+        tagged_metrics:
+          $ref: '#/components/schemas/EgressOnlyTracking.TaggedMetrics.Filter'
+          x-field-uid: 3
+    EgressOnlyTracking.TaggedMetrics.Filter:
+      description: Filter for tagged metrics
+      type: object
+      properties:
+        include:
+          description: Controls inclusion/exclusion of tagged metrics when fetching egress only tracking metrics.
+          type: boolean
+          default: true
+          x-field-uid: 1
+        include_empty_metrics:
+          description: Controls inclusion/exclusion of tagged metrics where each underlying attributes has zero value or absent value.
+          type: boolean
+          default: false
+          x-field-uid: 2
+        metric_names:
+          description: >-
+            The list of metric names that the returned result set will contain.
+            If the list is empty then all metrics will be returned.
+          type: array
+          items:
+            type: string
+            x-enum:
+              frames_tx:
+                x-field-uid: 1
+              frames_rx:
+                x-field-uid: 2
+              bytes_tx:
+                x-field-uid: 3
+              bytes_rx:
+                x-field-uid: 4
+              frames_tx_rate:
+                x-field-uid: 5
+              frames_rx_rate:
+                x-field-uid: 6
+              tx_l1_rate_bps:
+                x-field-uid: 7
+              rx_l1_rate_bps:
+                x-field-uid: 8
+              tx_rate_bytes:
+                x-field-uid: 9
+              rx_rate_bytes:
+                x-field-uid: 10
+              tx_rate_bps:
+                x-field-uid: 11
+              rx_rate_bps:
+                x-field-uid: 12
+              tx_rate_kbps:
+                x-field-uid: 13
+              rx_rate_kbps:
+                x-field-uid: 14
+              tx_rate_mbps:
+                x-field-uid: 15
+              rx_rate_mbps:
+                x-field-uid: 16
+          x-field-uid: 3
+        filters:
+          description: List of filters to selectively fetch tagged metrics with certain tag and corresponding value.
+          type: array
+          items:
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTag.Filter'
+          x-field-uid: 4
+    EgressOnlyTracking.MetricTag.Filter:
+      description: |-
+        A container for filtering ingress and/or egress metric tags.
+        The Tx stats may not be applicable in both the request and response filter.
+      type: object
+      properties:
+        name:
+          description: |-
+            A metric tag name that MUST exist in a packet or
+            egress_only_tracking configuration
+          type: string
+          x-field-uid: 1
+        values:
+          description: |-
+            A list of filters that can be applied to the metric tag name.
+            By default all values will be included in the egress-only-tracking metric results.
+          type: array
+          items:
+            type: string
+          x-field-uid: 2
+    EgressOnlyTracking.Metric:
+      description: |-
+        A container for egress-only-tracking metrics.
+        The container is keyed by the name, port_tx and port_rx.
+      type: object
+      properties:
+        name:
+          description: >-
+            The name of the egress-only-tracking flow
+          type: string
+          example: 'Tx -> Rx egress only'
+          x-field-uid: 1
+        port_tx:
+          description: |-
+            The name of the transmit port
+          type: string
+          x-field-uid: 2
+        port_rx:
+          description: |-
+            The name of the receive port
+          type: string
+          x-field-uid: 3
+        transmit:
+          description: >-
+            The transmit state of the flow.
+          type: string
+          x-field-uid: 5
+          x-enum:
+            started:
+              x-field-uid: 1
+            stopped:
+              x-field-uid: 2
+            paused:
+              x-field-uid: 3
+        frames_tx:
+          description: >-
+            The current total number of frames transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 6
+        frames_rx:
+          description: >-
+            The current total number of valid frames received
+          type: integer
+          format: uint64
+          x-field-uid: 7
+        bytes_tx:
+          description: >-
+            The current total number of bytes transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 8
+        bytes_rx:
+          description: >-
+            The current total number of bytes received
+          type: integer
+          format: uint64
+          x-field-uid: 9
+        frames_tx_rate:
+          description: >-
+            The current rate of frames transmitted
+          type: number
+          x-field-uid: 10
+        frames_rx_rate:
+          description: >-
+            The current rate of valid frames received
+          type: number
+          x-field-uid: 11
+        loss:
+          description: >-
+            The percentage of lost frames
+          type: number
+          x-field-uid: 12
+        timestamps:
+          $ref: '#/components/schemas/Metric.Timestamp'
+          x-field-uid: 13
+        latency:
+          $ref: '#/components/schemas/Metric.Latency'
+          x-field-uid: 14
+        tagged_metrics:
+          description: |-
+            List of metrics corresponding to a set of values applicable
+            for configured metric tags in ingress or egress packet header fields of corresponding flow.
+            The container is keyed by list of tag-value pairs.
+          type: array
+          items:
+            $ref: '#/components/schemas/EgressOnlyTracking.TaggedMetric'
+          x-field-uid: 15
+        tx_l1_rate_bps:
+          description: >-
+            The Layer 1 transmission rate in bits per second.
+          type: number
+          x-field-uid: 16
+        rx_l1_rate_bps:
+          description: >-
+            The Layer 1 receive rate in bits per second.
+          type: number
+          x-field-uid: 17
+        tx_rate_bytes:
+          description: >-
+            The transmission rate in bytes per second.
+          type: number
+          x-field-uid: 18
+        rx_rate_bytes:
+          description: >-
+            The receive rate in bytes per second.
+          type: number
+          x-field-uid: 19
+        tx_rate_bps:
+          description: >-
+            The transmission rate in bits per second.
+          type: number
+          x-field-uid: 20
+        rx_rate_bps:
+          description: >-
+            The receive rate in bits per second.
+          type: number
+          x-field-uid: 21
+        tx_rate_kbps:
+          description: >-
+            The transmission rate in Kilobits per second.
+          type: number
+          x-field-uid: 22
+        rx_rate_kbps:
+          description: >-
+            The receive rate in Kilobits per second.
+          type: number
+          x-field-uid: 23
+        tx_rate_mbps:
+          description: >-
+            The transmission rate in Megabits per second.
+          type: number
+          x-field-uid: 24
+        rx_rate_mbps:
+          description: >-
+            The receive rate in Megabits per second.
+          type: number
+          x-field-uid: 25
+    EgressOnlyTracking.TaggedMetric:
+      description: |-
+        Metrics for each set of values applicable for configured
+        metric tags in ingress or egress packet header fields of corresponding flow.
+        The container is keyed by list of tag-value pairs.
+      type: object
+      properties:
+        tags:
+          description: |-
+            List of tag and value pairs
+          type: array
+          items:
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTag'
+          x-field-uid: 1
+        frames_tx:
+          description: >-
+            The current total number of frames transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 2
+        frames_rx:
+          description: >-
+            The current total number of valid frames received
+          type: integer
+          format: uint64
+          x-field-uid: 3
+        bytes_tx:
+          description: >-
+            The current total number of bytes transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 4
+        bytes_rx:
+          description: >-
+            The current total number of bytes received
+          type: integer
+          format: uint64
+          x-field-uid: 5
+        frames_tx_rate:
+          description: >-
+            The current rate of frames transmitted
+          type: number
+          x-field-uid: 6
+        frames_rx_rate:
+          description: >-
+            The current rate of valid frames received
+          type: number
+          x-field-uid: 7
+        loss:
+          description: >-
+            The percentage of lost frames
+          type: number
+          x-field-uid: 8
+        timestamps:
+          $ref: '#/components/schemas/Metric.Timestamp'
+          x-field-uid: 9
+        latency:
+          $ref: '#/components/schemas/Metric.Latency'
+          x-field-uid: 10
+        tx_l1_rate_bps:
+          description: >-
+            The Layer 1 transmission rate in bits per second.
+          type: number
+          x-field-uid: 11
+        rx_l1_rate_bps:
+          description: >-
+            The Layer 1 receive rate in bits per second.
+          type: number
+          x-field-uid: 12
+        tx_rate_bytes:
+          description: >-
+            The transmission rate in bytes per second.
+          type: number
+          x-field-uid: 13
+        rx_rate_bytes:
+          description: >-
+            The receive rate in bytes per second.
+          type: number
+          x-field-uid: 14
+        tx_rate_bps:
+          description: >-
+            The transmission rate in bits per second.
+          type: number
+          x-field-uid: 15
+        rx_rate_bps:
+          description: >-
+            The receive rate in bits per second.
+          type: number
+          x-field-uid: 16
+        tx_rate_kbps:
+          description: >-
+            The transmission rate in Kilobits per second.
+          type: number
+          x-field-uid: 17
+        rx_rate_kbps:
+          description: >-
+            The receive rate in Kilobits per second.
+          type: number
+          x-field-uid: 18
+        tx_rate_mbps:
+          description: >-
+            The transmission rate in Megabits per second.
+          type: number
+          x-field-uid: 19
+        rx_rate_mbps:
+          description: >-
+            The receive rate in Megabits per second.
+          type: number
+          x-field-uid: 20
+    EgressOnlyTracking.MetricTag:
+      type: object
+      properties:
+        name:
+          description: |-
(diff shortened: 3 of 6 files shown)


indraniBh:
FYI @arkajyoti-cloud @sasubrata @alakendu

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
+components:
+  schemas:
+    EgressOnlyTacking:
```
apratimmukherjee, line 3: Tracking

config/config.yaml
```diff
@@ -64,6 +64,13 @@ components:
           items:
             $ref: '../lldp/lldp.yaml#/components/schemas/Lldp'
           x-field-uid: 9
+        egress_only_tracking:
+          description: |-
+            The list of structures containing Egress offset and it's length to be placed in the packet.
```
apratimmukherjee, line 69: "Container for information on tracking desired based on certain offset and length of bits  on ports for received packets on specified receive ports. It enables the user to retrieve information based on number of unique packets received for each unique value in the bits being tracked from the beginning of the test ." --> Something of this nature might be more helpful to the user. The most appropriate description needs probably more detailed understanding but this is what I understand at a high level. [ But definitely nothing being placed in the packet ]

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+      description: |-
+        The list of structures containing Egress offset and it's length to be placed in the packet.
```
apratimmukherjee, line 5: "One specific egress tracking specification for a  specified port. An application which supports a single or limited number of egress tracking specifications for a single port should return an error if the number of egress tracking specifications for a specific port or in total exceeds its capabilities. "  --> Note : All these suggestions are based on my high level understanding and should be adjusted based on what the user can expect from this feature.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+      type: object
+      required: [port_name]
+      properties:
+        port_name:
+          description: Name of the Rx port where egress only tracking enabled.
```
apratimmukherjee, line 10: Name of the received port where this egress tracking rule/specification has to be applied.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+            One or more metric tags can be used to enable tracking portion of or all bits
+            in a corresponding header field for metrics per each applicable value.
+            These would appear as tagged metrics in corresponding egress_only_tracking metrics.
+          type: array
+          items: 
+            $ref: '#/components/schemas/EgressOnlyTacking.MetricTag'
+          x-field-uid: 2
+        filter:
+          description: |-
+            Offset in bits where the tracking value will be placed in the packet.
```
apratimmukherjee, line 26: "Specified a rule which will be used to first filter received packets on this port before applying the egress tracking metric_tags on them. If the packet does not match the rule, it will not be considered for egress tracking."  -> Again might not more accuracy but definitely current description is not appropriate.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+        Metric Tag can be used to enable tracking portion of or all bits
+        in a corresponding header field for metrics per each applicable value.
+        These would appear as tagged metrics in corresponding flow metrics.
+      type: object
+      required:
+      - name
+      properties:
+        name:
+          description: |-
+            Name used to identify the metrics associated with the values applicable
```
apratimmukherjee, line 40: Unique name ?

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+            for configured offset and length inside corresponding header field
+          type: string
+          pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
+          x-field-uid: 1
+        offset:
+          description: |-
+            Offset in bits relative to start of corresponding header field
```
apratimmukherjee, line 47: corresponding header field or start of packet . I dont see any packet shape in this proposal to be able to specify in terms of headers. Not sure if we should have something like that though.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+          type: integer
+          format: uint32
+          default: 0
+          x-field-uid: 2
+        length:
+          description: |-
+            Number of bits to track for metrics starting from configured offset
+            of corresponding header field
```
apratimmukherjee, line 55: from start of packet ?

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,74 @@
...
+          format: uint32
+          default: 1
+          minimum: 1
+          x-field-uid: 3
+    EgressOnlyTacking.Filter:
+      type: object
+      properties:
+        choice:
+          description: >-
+            Offset in bits where the tracking value will be placed in the packet.
```
apratimmukherjee, line 66: Specification of a rule to filter incoming packets on the specified port. If a packet does not match the filter it will not be considered for egress tracking. 
Currently two options are provided: 
none: All packets will be considered for egress only tracking. 
auto.macsec:  This requires that macsec enabled Ethernet interface should be configured on this port. This filter will ensure that only packets with Ethernet Type set to MacSec ( 0xabcd) will be considered for egress only tracking and other packets will be filtered out for the purpose of egress only tracking.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+        port_names:
+          description: |-
+            Egress only tracking metrics will be retrieved for these port names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - "/components/schemas/EgressOnlyTacking/properties/port_name"
+          x-field-uid: 1
+        metric_names:
```
apratimmukherjee, line 17: Why do we have this field. Note there is no per flow stats anymore.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+          description: |-
+            The name of the transmit port
+          type: string
+          x-field-uid: 2
+        port_rx:
+          description: |-
+            The name of the receive port
+          type: string
+          x-field-uid: 3
+        transmit:
```
apratimmukherjee, line 166: Does not make sense

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+          items:
+            type: string
+          x-field-uid: 2
+    EgressOnlyTracking.Metric:
+      description: |-
+        A container for egress-only-tracking metrics.
+        The container is keyed by the name, port_tx and port_rx.
+      type: object
+      properties:
+        name:
```
apratimmukherjee, line 150: I think the key will probably be rx port name and metric name , maybe some tracking info i.e. what is the tracking offset and bit , value of bits and then the rx stats for matched packets i.e frames/rate etc. as makes sense . Current looks wrong. It will not be same as flow egress tracking . Please revisit the stats part completely. Check IxN once.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+            The name of the transmit port
+          type: string
+          x-field-uid: 2
+        port_rx:
+          description: |-
+            The name of the receive port
+          type: string
+          x-field-uid: 3
+        transmit:
+          description: >-
```
apratimmukherjee, line 167: Nothing being transmitted. No state.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+          type: string
+          x-field-uid: 5
+          x-enum:
+            started:
+              x-field-uid: 1
+            stopped:
+              x-field-uid: 2
+            paused:
+              x-field-uid: 3
+        frames_tx:
```
apratimmukherjee, line 178: No tx

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+          x-field-uid: 11
+        loss:
+          description: >-
+            The percentage of lost frames
+          type: number
+          x-field-uid: 12
+        timestamps:
+          $ref: '#/components/schemas/Metric.Timestamp'
+          x-field-uid: 13
+        latency:
```
apratimmukherjee, line 220: No latency

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,471 @@
...
+          description: >-
+            The current rate of frames transmitted
+          type: number
+          x-field-uid: 10
+        frames_rx_rate:
+          description: >-
+            The current rate of valid frames received
+          type: number
+          x-field-uid: 11
+        loss:
```
apratimmukherjee, line 212: No loss

apratimmukherjee (changes requested):
The metrics response looks like needs major changes. Config side mainly descriptions need updates. Given some example test but might not be fully accurate.

egress/egress-only-tracking.yaml
indraniBh: done.

egress/egress-only-tracking.yaml
indraniBh: Updated.

egress/egress-only-tracking.yaml
indraniBh: Updated.

egress/egress-only-tracking.yaml
indraniBh: done.

egress/egress-only-tracking.yaml
indraniBh: I make it "start of the packet"

egress/egress-only-tracking.yaml
indraniBh: done.

egress/egress-only-tracking.yaml
indraniBh: Updated.

result/egressonlytracking.yaml
indraniBh: Done.

result/egressonlytracking.yaml
indraniBh: I am keeping only name and port_rx. @sasubrata Can you please add whether we can add more tracking info as key in the description.

result/egressonlytracking.yaml
indraniBh: removed

result/egressonlytracking.yaml
indraniBh: Removed.

result/egressonlytracking.yaml
indraniBh: I have removed all tx related stats.

result/egressonlytracking.yaml
indraniBh: Removed.

result/egressonlytracking.yaml
indraniBh: Removed.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name
+      properties:
+        name:
+          description: |-
+            The Name used to identify the metrics associated with the values applicable
+            for configured offset and length inside corresponding header field.
+          type: string
+          pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
+          x-field-uid: 1
+        offset:
```
PrasenjitAdhikary, line 49: Redocly view is showing a different hierarchy - having name and value (with subfields of choice, hex, str). Might be some issue in generated code.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,84 @@
...
+            Specified a rule which will be used to first filter received packets on this port 
+            before applying the egress tracking metric_tags on them. 
+            If the packet does not match the rule, it will not be considered for egress tracking.
+          $ref: '#/components/schemas/EgressOnlyTracking.Filter'
+          x-field-uid: 3
+    EgressOnlyTracking.MetricTag:
+      description: |-
+        Metric Tag can be used to enable tracking portion of or all bits
+        in a corresponding header field for metrics per each applicable value.
+        These would appear as tagged metrics in corresponding flow metrics.
```
PrasenjitAdhikary, line 37: The result is not part of corresponding flow metrics. It will be shown in new "egress_only_tracking" metrics

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+          x-field-uid: 2
+        metric_names:
+          description: >-
+            The list of metric names that the returned result set will contain.
+            If the list is empty then all metrics will be returned.
+          type: array
+          items:
+            type: string
+            x-enum:
+              frames_tx:
```
apratimmukherjee, line 42: How will you know frames_tx ?

apratimmukherjee: Need egress only tracking stats here , not flow egress tracking.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+          description: |-
+            Egress only tracking metrics will be retrieved for these port names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - "/components/schemas/EgressOnlyTracking/properties/port_name"
+          x-field-uid: 1
+        tagged_metrics:
+          $ref: '#/components/schemas/EgressOnlyTracking.TaggedMetrics.Filter'
```
apratimmukherjee, line 18: Cant we put the fields directly here ? What is the advantage of a separate class ?

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+        filters:
+          description: List of filters to selectively fetch tagged metrics with certain tag and corresponding value.
+          type: array
+          items:
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTag.Filter'
+          x-field-uid: 4
+    EgressOnlyTracking.MetricTag.Filter:
+      description: |-
+        A container for filtering ingress and/or egress metric tags.
+        The Tx stats may not be applicable in both the request and response filter.
```
apratimmukherjee, line 84: Why ingress ? Why Tx stats ?

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+        The container is keyed by the name and port_rx.
+      type: object
+      properties:
+        name:
+          description: >-
+            The name of the egress-only-tracking flow
+          type: string
+          example: 'Tx -> Rx egress only'
+          x-field-uid: 1
+        port_tx:
```
apratimmukherjee, line 113: Need to understand if port_tx is really known. If yes, should clearly mention that this should be included only if implementation is able to detect if the received packet was transmitted by one of the ports in the configuration in set_config , otherwise it should not be included. Also, key fields should be one top. Not this.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+              rx_rate_mbps:
+                x-field-uid: 16
+          x-field-uid: 3
+        filters:
+          description: List of filters to selectively fetch tagged metrics with certain tag and corresponding value.
+          type: array
+          items:
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTag.Filter'
+          x-field-uid: 4
+    EgressOnlyTracking.MetricTag.Filter:
```
apratimmukherjee, line 81: Why do we have both Metric and Tagged Metric. Isn't this PR for egress only tracking ?

apratimmukherjee (changes requested):
Second round of review.

egress/egress-only-tracking.yaml
indraniBh: Fixed.

egress/egress-only-tracking.yaml
indraniBh: done.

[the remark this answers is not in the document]
indraniBh: done.

config/config.yaml
```diff
@@ -64,6 +64,16 @@ components:
           items:
             $ref: '../lldp/lldp.yaml#/components/schemas/Lldp'
           x-field-uid: 9
+        egress_only_tracking:
+          description: |-
+            Container for information on tracking desired based on certain offset and length of bits on ports 
+            for received packets on specified receive ports. 
+            It enables the user to retrieve information based on number of unique packets 
+            received for each unique value in the bits being tracked from the beginning of the test .
```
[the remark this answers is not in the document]
indraniBh: @PrasenjitAdhikary , @sasubrata Can you please suggest/check the description for the above.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,84 @@
...
+        metric_tags:
+          description: |-
+            One or more metric tags can be used to enable tracking portion of or all bits
+            in a corresponding header field for metrics per each applicable value.
+            These would appear as tagged metrics in corresponding egress_only_tracking metrics.
+          type: array
+          items: 
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTag'
+          x-field-uid: 2
+        filter:
```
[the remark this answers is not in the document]
indraniBh: Ok.

egress/egress-only-tracking.yaml
```diff
@@ -0,0 +1,84 @@
...
+      type: object
+      properties:
+        choice:
+          description: >-
+            Specification of a rule to filter incoming packets on the specified port. 
+            If a packet does not match the filter it will not be considered for egress tracking.
+            Currently two options are provided:
+            none: All packets will be considered for egress only tracking.
+            auto.macsec: This requires that macsec enabled Ethernet interface should be configured on this port. 
+            This filter will ensure that only packets with Ethernet Type set to MacSec ( 0xabcd) will be considered for egress only tracking
```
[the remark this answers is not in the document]
indraniBh: Changed to 0x88E5 (ether type of macsec)

result/egressonlytracking.yaml
indraniBh: Removed tagged metric.

sasubrata (commented):
Adding review comments. We can proceed with current comments and rework. The model can be modified along with implementation/ testing.

config/config.yaml
```diff
@@ -64,6 +64,16 @@ components:
           items:
             $ref: '../lldp/lldp.yaml#/components/schemas/Lldp'
           x-field-uid: 9
+        egress_only_tracking:
```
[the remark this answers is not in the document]
anjan-keysight: I agree, we may have tracking based on frame size, so a generic node like "other-tracking-options" should be considered. Egress-only-tracking, frame-size based tracking can be sub-nodes under it.

config/config.yaml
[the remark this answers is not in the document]
anjan-keysight: Rephase the first part of description. "Container for information on desired tracking, based on an offset and length of bits..."

egress/egress-only-tracking.yaml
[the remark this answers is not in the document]
anjan-keysight: Those filter values are "orthogonal". We should report stats for each value.

result/egressonlytracking.yaml
anjan-keysight: The metrics name and value (enclosed in a "tags" node), and corresponding stats should an array inside rx_port. So you will have an array of rx port (top level); each rx_port will contain an array of a tag node (with metrics name and value) and corresponding stats. Refer to egress tracking examples.

config/config.yaml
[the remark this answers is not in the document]
sasubrata: "tracking_options" looks good to me. "egress_only" can be a sub node of it.

egress/egress-only-tracking.yaml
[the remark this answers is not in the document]
sasubrata: @anjan-keysight filter is similar to signature to match in incoming packets. Once packets are filtered, egress only tracking is applied. Are you saying metric tags are orthogonal i.e. reports stats for each tag? I have given similar comment below.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,290 @@
...
+        port_names:
+          description: |-
+            Egress only tracking metrics will be retrieved for these port names.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - "/components/schemas/EgressOnlyTracking/properties/port_name"
+          x-field-uid: 1
+        tagged_metrics:
```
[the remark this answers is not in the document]
apratimmukherjee: Resolved with change of design.

config/config.yaml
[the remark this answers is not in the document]
apratimmukherjee: As per discussion, if future need comes, this can be handled via new option predefined_tags construct similar to flow egress tracking.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,250 @@
+components:
+  schemas:
+    EgressOnlyTracking.Metrics.Request:
+      description: >-
+        The container for a egress only tracking metric request.
+      type: object
+      properties:
+        port_names:
+          description: |-
+            Egress only tracking metrics will be retrieved for these port names.
```
apratimmukherjee, line 10: Maybe we should add . If no port_names are provided , egress_only_tracking stats will be returned for all ports which have one or more egress_only_tracking enabled .

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,250 @@
...
+          x-field-uid: 1
+        tagged_metrics:
+          $ref: '#/components/schemas/EgressOnlyTracking.TaggedMetrics.Filter'
+          x-field-uid: 2
+    EgressOnlyTracking.TaggedMetrics.Filter:
+      description: Filter for tagged metrics
+      type: object
+      properties:
+        include_empty_metrics:
+          description: Controls inclusion/exclusion of tagged metrics where each underlying attributes has zero value or absent value.
```
apratimmukherjee, line 25: attribute

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,250 @@
...
+      type: object
+      properties:
+        choice:
+          description: Available formats for metric tag value
+          type: string
+          default: hex
+          x-enum:
+            hex:
+              x-field-uid: 1
+            str:
```
apratimmukherjee, line 162: @PrasenjitAdhikary @anjan-keysight  : I see it is similar for flow egress tracking. But it is not clear when the implementation should return as hex and when as string. Should it not always be in hex ?  Will not different implementations possibly return data in different format if this choice is kept ?

apratimmukherjee (approved):
In general, looks good to me. 
1.  Some typos and better description suggested. 
2.  More information on expected response for egress_only_tracking stats fetch , say with multiple tracking or at least the mac sec use-case. 
3. Doubt needs to be resolved on value choice as hex/str in returned stats as to when an implementation will switch to non default str format.

result/egressonlytracking.yaml
indraniBh: Added.

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+            One or more metric tags can be used to enable tracking portion of or all bits
+            in a corresponding header field for metrics per each applicable value.
+            These would appear as tagged metrics in corresponding egress_only_tracking metrics.
+          type: array
+          items: 
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTags'
+          x-field-uid: 2
+        filter:
+          description: |-
+            Specified a rule which will be used to first filter received packets on this port
```
sasubrata, line 28: Specified => Specifies

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+    EgressOnlyTracking.Filter:
+      type: object
+      properties:
+        choice:
+          description: >- 
+            If a packet does not match the filter it will not be considered for egress tracking.
+            Currently two options are provided:
+            none: All packets will be considered for egress only tracking.
+            auto.macsec: This requires that macsec enabled Ethernet interface should be configured on this port. 
+            This filter will ensure that only packets with Ethernet Type set to MacSec ( 0x88E5) will be considered for egress only tracking
```
sasubrata, line 77: MacSec => MACsec

sasubrata, line 77: Please note the filter also checks if destination MAC address belongs to traffic Rx device(s):

Accordingly update description:

This filter will ensure that only packets with Ethernet Type set to MACsec (0x88E5) and destined to traffic Rx device(s) will be considered for egress only tracking

sasubrata (commented):
Some more review comments on config part.

config/config.yaml
sasubrata, line 67: As this field is an array, shall we make the name plural "egress_only_trackings" to indicate that is an array? This is the nomenclature I saw in OTG model.

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+        metric_tags:
+          description: |-
+            One or more metric tags can be used to enable tracking portion of or all bits
+            in a corresponding header field for metrics per each applicable value.
+            These would appear as tagged metrics in corresponding egress_only_tracking metrics.
+          type: array
+          items: 
+            $ref: '#/components/schemas/EgressOnlyTracking.MetricTags'
+          x-field-uid: 2
+        filter:
```
sasubrata, line 26: As this field is an array, shall we make the name plural "filters" to indicate that is an array? This is the nomenclature I saw in OTG model.

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+          pattern: ^[\sa-zA-Z0-9-_()><\[\]]+$
+          x-field-uid: 1
+        offset:
+          description: |-
+            Offset in bits relative to start of the packet.
+          type: integer
+          format: uint32
+          x-field-uid: 2
+        length:
+          description: |-
```
sasubrata, line 60: Just wanted to add that one metric tag (offset, length) combination allows only tracking consecutive bits. It does not allow non-consecutive bits.

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+            auto_macsec: This requires that MACsec enabled Ethernet interface should be configured on this port. 
+            This filter will ensure that only packets with Ethernet Type set to MACsec (0x88E5) and 
+            destined to traffic Rx device(s) will be considered for egress only tracking
+          type: string
+          x-field-uid: 1
+          default: none
+          x-enum:
+            none:
+              x-field-uid: 1
+            auto_macsec:
```
sasubrata, line 85: Can we make it simply "auto"? As MACsec protocol implementation is supposed to set the filter internally, another new protocol can do the same. "auto" setting will be applicable for all protocols.

Please comment.

egress/egressonlytracking.yaml
apratimmukherjee: Yes, we can add multiple tags for such use-case to support non-consecutive bits . So use-case should still be supported, right ?

egress/egressonlytracking.yaml
apratimmukherjee: Ideally would not want to have auto at all, it is there to be able to map to the specific MAC Sec requirements for snapp_ixn . If auto is there at all for some other protocol, meaning would probably be very different and ideally avoided completely with generic custom option. But easier to add a generic auto later if must have rather than get stuck with an auto that needs to do something different for some non mac-sec usecase. So prefer having macsec specified specifically here .

sasubrata: This egress_only_tracking filter will be valid for paths other than snappi-ixnetwork, right? For ixia-c too?

"auto" means that user need not understand exactly what is set internally for any protocol but she knows it will work - for any protocol. In other words, she does not need to understand the difference between "auto_macsec" and "auto_protocol_x" as long as the respective protocol packets are filtered in.

Having separate choices per protocol e.g. "auto_macsec" and "auto_protocol_x" adds to maintainability and user has to set it based on protocol under test.

Another thing. Let's say protocol MACsec  and protocol "x" can be stacked in same packet, then it the burden is on the user to choose what to set: auto_macsec or auto_protocol_x. Ideally the implementation should inspect the stack and very likely set the outer protocol as filter. If there is a single "auto" choice, user need not bother.

If "auto" is added later, internally its setting will be per protocol.

Am I missing something?

apratimmukherjee: Yes, the model is meant to be generic to handle other Keysight impl. like ixia-c or otg-hw but also non Keysight implementation such as lemming. 
auto_protocol_x/y allows
1) To clearly specify that choosing the auto is meant to do what in the backend / in terms of behavior
2) If **_BOTH_** protocol x and y are there in the config, which auto mode to choose . 
So looks much more generic to me than just auto. ( similar to auto.config vs auto.dhcp for src/dst IP in flows )

egress/egressonlytracking.yaml
sasubrata: Yes, it can be supported. But breaking one tag into multiple tags may exhaust number of total egress trackers supported by an implementation. E.g. IxNetwork supports maximum 3 trackers.

I do not want to make the model unnecessarily heavy. Just wanted us to know the limitation.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,252 @@
...
+            The name of the transmit port
+          type: string
+          x-field-uid: 1
+        frames_tx:
+          description: >-
+            The current total number of frames transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 2
+        bytes_tx:
```
sasubrata, line 211: Bytes Tx: This stat is not published in egress only flow statistics. This stat should be removed from supported Tx metrics.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,252 @@
...
+            The current total number of bytes transmitted
+          type: integer
+          format: uint64
+          x-field-uid: 3
+        frames_tx_rate:
+          description: >-
+            The current rate of frames transmitted
+          type: number
+          x-field-uid: 4
+        loss:
```
sasubrata, line 222: Loss is a an Rx property. Please move it out of Tx metrics and move in main (Rx) metrics.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,252 @@
...
+          x-field-uid: 3
+    EgressOnlyTracking.Timestamp:
+      description: |-
+        The container for timestamp metrics.
+        The container will be empty if the timestamp has not been configured for the flow.
+      type: object
+      properties:
+        first_timestamp_ns:
+          description: |-
+            First timestamp in nanoseconds
```
sasubrata, line 184: IxNetwork provides timestamp in milliseconds granularity i.e. in hh:mm:ss.mmm i.e. hour:minute:second:milliseconds. Will convert it to nanoseconds.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,252 @@
...
+          type: number
+          format: double
+          x-field-uid: 1
+        last_timestamp_ns:
+          description: |-
+            Last timestamp in nanoseconds
```
sasubrata, line 190: IxNetwork provides timestamp in milliseconds granularity i.e. in hh:mm:ss.mmm i.e. hour:minute:second:milliseconds. Will convert it to nanoseconds.

result/egressonlytracking.yaml
sasubrata: It should always return either hex or integer (only one type) as the value will always be numeric. In that case, can we keep value as just "type: integer" as integer format will be user friendly and this is the format user has configured protocol field while configuring e.g. VLAN 10.

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,252 @@
...
+            The current total number of bytes received
+          type: integer
+          format: uint64
+          x-field-uid: 3
+        frames_rx_rate:
+          description: >-
+            The current rate of valid frames received
+          type: number
+          x-field-uid: 4
+        timestamps:
```
sasubrata, line 110: How to indicate from script whether timestamps are desired or not? In case of regular flow metrics, it is indicated from of flow[]->metrics->timestamps (boolean) configuration.

result/egressonlytracking.yaml
apratimmukherjee: @sasubrata  It is debatable : I think for some fields such as MAC Hex makes more sense and might be for some decimal makes more sense. But if I would look for a generic option, I would always prefer hex . Seems to make more sense for random bytes in the packet and aligns with wireshark as well.  My doubt for @PrasenjitAdhikary @anjan-keysight was more on the lines of why two choices are here and when will which choice be used.

config/config.yaml
apratimmukherjee: @sasubrata @indraniBh : Normally the 's' at the end is fine but for this particular case I feel egress_only_trackings sounds little peculiar though fine with either. I think egress_only_tracking[i] easily indicates that this is an array without soundinf odd.

result/egressonlytracking.yaml
apratimmukherjee: @sasubrata : I dont think we can do that. The reason it is kept here is that loss can be calculated only if Tx is known . So if Tx itself is not known , which would be the generic case for egress tracking ; no impl, can even provide the loss , so I think it is better kept as part of the optional group of   stats denoted by tx_metrics ; since we dont want to probably add even more optional fields.

result/egressonlytracking.yaml
apratimmukherjee: Yes, agree if at all this is needed, there should be some way in egress only tracking config whether timestamps are needed or not, similar to flows.

sasubrata: As discussed, enable_timestamps (default = false) will be added per element of egress_only_tracking array in set_config

result/egressonlytracking.yaml
sasubrata: Ok, agreed

result/egressonlytracking.yaml
```diff
@@ -0,0 +1,230 @@
...
+                x-field-uid: 4
+              rx_rate_bytes:
+                x-field-uid: 5
+              rx_rate_bps:
+                x-field-uid: 6
+              rx_rate_kbps:
+                x-field-uid: 7
+              rx_rate_mbps:
+                x-field-uid: 8
+              tx_metrics:
```
sasubrata, line 56: If "tx_metrics" is present in desired metric_names, all supported Tx related metrics will be returned.

egress/egressonlytracking.yaml
```diff
@@ -0,0 +1,86 @@
...
+        should return an error if the number of egress tracking specifications for a specific port or in total exceeds its capabilities.
+      type: object
+      required: [port_name]
+      properties:
+        port_name:
+          description: Name of the received port this egress tracking rule/specification has to be applied.
+          type: string
+          x-constraint:
+          - '/components/schemas/Port/properties/name'
+          x-field-uid: 1
```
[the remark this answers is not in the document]
sasubrata: aptratimmukherjee added:

Could user not do it simply like 

               port1
                              Vlan100.port1
 
               Port2
                              Vlan200.port2 or whatever. 
 
Etc. 
Seems okay to me at first glance. Same is true for devices or interfaces or anything else across multiple ports right ?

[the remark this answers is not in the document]
sasubrata: Per port example in mail below looks ok to me. Only thing is that identical config except the tag name is repeated across ports. But max number of ports is limited - in single or two digits on the lower side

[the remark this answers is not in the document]
sasubrata: anjan-keysight added:

For similar tracking type (and all other tracking parameters being same) across multiple ports we would like to keep the metrics tag name same. It’s easier for the user to process. That was the idea.

But on second thought, we can still implement that in current model. Here we can have same metrics tag name occur multiple times for similar tracking type across multiple ports.

Subrata, you may ignore that comment then.

PR #418, opened 2025-03-25
MACsec flow header

sasubrata:
Add MACsec header in flow so that handle to MACsec payload e.g. IPv4 can be obtained 
 by flow.packet.ethernet().macsec().ipv4() header sequence. This is required to shape MACsec payload e.g. IPv4 DSCP properties. As of now MACsec header choice is only "auto" i.e. it is automatically set based on MACsec protocol configuration and it cannot be overwritten from flow. In future, we may add support for overwriting MACsec header fields from flow.


flow/packet-headers/macsec.yaml
```diff
@@ -0,0 +1,14 @@
+components:
+  schemas:
+    Flow.Macsec:
+      description: MACsec packet header.
+      type: object
+      properties:
+        choice:
+          description: Currently only auto choice is allowed. If choice is auto, MACsec header is autogenerated. In future, custom choice will be given to set header fields.
```
apratimmukherjee, line 8: 1) I think we can mention that if auto choice is selected, flow.tx_rx.choice must be of type 'device' and flow.tx_rx.device.tx_names[0] must be chosen to be an endpoint that is on or behind a MACSec enabled ethernet to be able to correctly auto-fill the fields of the MACsec header . If either of this is not true, the implementation should return an error specifying the issue . 
2) Maybe : "a custom choice can be added to allow user to control specific MACsec header fields or/and to generate flow.tx_rx.port type of traffic with MACSec header with all fields explicitly specified by the user."

apratimmukherjee (approved):
Looks good in general. 
Might be useful to enhance some of the documentation for the auto choice in terms of when it is usable and when it will return error as well as some more clarification on extendibility with custom type in the future.

flow/packet-headers/macsec.yaml
sasubrata: I have updated description with above information.

PrasenjitAdhikary:
LGTM

PrasenjitAdhikary (approved):
LGTM

PR #427, opened 2025-07-07
BGP Monitoring Protocol : Client support. ( Approach1) 

apratimmukherjee:
WIP

Redocli : https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/dev-egress-only2/artifacts/openapi.yaml&nocors#tag/Configuration/operation/set_config

### DIFF
--- a/result/bmpprefix.yaml
+++ b/result/bmpprefix.yaml
@@ -1,0 +1,639 @@
+components:
+  schemas:
+    BmpPrefix.State.Request:
+      description: >-
+        The request to retrieve BMP server learned BGP peer prefix information.
+      type: object
+      properties:
+        peer_bgp_id:
+          description: >-
+            The identifiers of BGP peers for which prefix information will be
+            retrieved. If no names are specified then the results will
+            contain prefix information for all BGP peers in BMP server database.
+          type: array
+          items:
+            type: string
+          x-field-uid: 1
+        prefix_filters:
+          description: >-
+            Specify which prefixes to return. If the list is empty or missing
+            then all prefixes will be returned.
+          type: array
+          items:
+            type: string
+            x-enum:
+              ipv4_unicast:
+                x-field-uid: 1
+              ipv6_unicast:
+                x-field-uid: 2
+          x-field-uid: 2
+        ipv4_unicast_filters:
+          description: >-
+            The IPv4 unicast results can be filtered by specifying additional
+            prefix search criteria. If the ipv4_unicast_filters property is
+            missing or empty then all IPv4 prefixes will be returned.
+          type: array
+          items:
+            $ref: '#/components/schemas/BmpPrefix.Ipv4Unicast.Filter'
+          x-field-uid: 3
+        ipv6_unicast_filters:
+          description: >-
+            The IPv6 unicast results can be filtered by specifying additional
+            prefix search criteria. If the ipv6_unicast_filters property is
+            missing or empty then all IPv6 prefixes will be returned.
+          type: array
+          items:
+            $ref: '#/components/schemas/BmpPrefix.Ipv6Unicast.Filter'
+          x-field-uid: 4
+    BmpPrefix.Ipv4Unicast.Filter:
+      type: object
+      properties:
+        addresses:
+          description: >-
+            The addresses to match. If the addresses property is missing or
+            empty then all addresses will match.
+          type: array
+          items:
+            type: string
+            format: ipv4
+          x-field-uid: 1
+        prefix_length:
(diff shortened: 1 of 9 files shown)


result/bmpserver.yaml
```diff
@@ -0,0 +1,129 @@
...
+            Information TLV in Initiation message.
+          type: string
+          x-field-uid: 3
+        system_description:
+          description: >-
+            The Information sysDesc MIB-II [RFC1213] object information carried in Information field of 
+            Information TLV in Initiation message.
+          type: string
+          x-field-uid: 4
+        termination_string:
```
apratimmukherjee, line 49: Should we have these within a state = up/down and down might have these reasons while up has info per peers ? 
Or maybe return information for up peers at that time , like BGP ?

result/bmpserver.yaml
```diff
@@ -0,0 +1,129 @@
...
+            The remote port number associated with the peering TCP session, or 0 if no TCP session actually exists.
+          type: integer
+          format: uint32
+          x-field-uid: 4
+        peer_bgp_id:
+          description: >-
+            The BGP Identifier of the peer.
+          type: string
+          x-field-uid: 5
+        peer_down_reason:
```
apratimmukherjee, line 113: Again , should this be within some current state and if down then show reason and if up show per peer info ?

result/bmpprefix.yaml
```diff
@@ -0,0 +1,652 @@
+components:
+  schemas:
+    BmpPrefix.State.Request:
+      description: >-
+        The request to retrieve BMP server learned BGP peer prefix information.
+      type: object
+      properties:
+        peer_bgp_id:
```
apratimmukherjee, line 8: What if two BMP servers are connected to same BGP peer ... this probably needs to have BMP server name and then array of peers with route info.

result/bmpprefix.yaml
```diff
@@ -0,0 +1,652 @@
...
+          description: >-
+            The identifiers of BGP peers for which prefix information will be
+            retrieved. If no names are specified then the results will
+            contain prefix information for all BGP peers in BMP server database.
+          type: array
+          items:
+            type: string
+          x-field-uid: 1
+        prefix_filters:
```
apratimmukherjee, line 17: I see it is there in BGP but confusing and best avoided I feel ( from legacy initial modeling ; best avoided I think )

result/bmpprefix.yaml
```diff
@@ -0,0 +1,652 @@
+components:
+  schemas:
+    BmpPrefix.State.Request:
+      description: >-
+        The request to retrieve BMP server learned BGP peer prefix information.
+      type: object
+      properties:
```
apratimmukherjee, line 7: I think array of BMP server names is a must , anything else would be maybe per BMP server name if at all.. not sure we need more depth of control

result/bmpprefix.yaml
```diff
@@ -0,0 +1,652 @@
...
+          type: array
+          items:
+            type: string
+            x-enum:
+              ipv4_unicast:
+                x-field-uid: 1
+              ipv6_unicast:
+                x-field-uid: 2
+          x-field-uid: 2
+        ipv4_unicast_filters:
```
apratimmukherjee, line 30: These are definitely overkill

result/bmpprefix.yaml
```diff
@@ -0,0 +1,652 @@
...
+      description: >-
+        BMP server learned bgp peer prefixes.
+      type: object
+      properties:
+        peer_bgp_id:
+          description: >-
+            The Identifier of a BGP peer.
+          type: string
+          x-field-uid: 1
+        ipv4_unicast_prefixes:
```
apratimmukherjee, line 144: I think a pre-policy and post-policy view of routes would be more useful [ within BMP server --> per server per peer as mentioned ]

apratimmukherjee (commented):
Initial structure based review comments.

result/bmpserver.yaml
```diff
@@ -0,0 +1,832 @@
...
+      properties:
+        bmp_server_names:
+          description: >-
+            The names of the BMP Servers  to return results for.
+            An empty list will return results for all BMP Servers.
+          type: array
+          items:
+            type: string
+          x-constraint:
+          - "/components/schemas/Device.Bmp.Server/properties/name"
```
alakendu: I think this `x-constraint` could be similar as `BmpServer.Metrics.Request` e,g, 
```
          - "/components/schemas/Device.Bmp.ServerV4/properties/name"
          - "/components/schemas/Device.Bmp.ServerV6/properties/name"
```

apratimmukherjee: Agreed. Fixed.

PR #430, opened 2025-07-22
[OSPFv2 ]Fix default/max of ospfv2/interface/advanced/routing_metric

rudranil-das:
Following fixes in ospfv2/interface/advanced/routing_metric are done for in this PR,
- introduce max value of 65535
  - As in RFC, length of router-LSA (Type-1) is 16-bit

- change default value from 0 to 10
  - It should have non-zero value reflecting cost of metric as default.


SuryyaKrJana (approved):
looks fine

PR #432, opened 2025-08-07
Addition of payload configurations in Flow

rudranil-das:


flow/payload.yaml
```diff
@@ -0,0 +1,47 @@
+components:
+  schemas:
+    Flow.Payload:
+      description: |-
+        A container for different types of payload, which is 
+        the data in the frame after protocol headers and instrumentation bytes.
```
PrasenjitAdhikary, line 6: Need to correct the description.
Payload data will start immediately after protocol header, not after instrumentation bytes. Some part of the payload will be overwritten with instrumentation data. Contents of instrumentation data and placement is implementation specific.

rudranil-das: updated, Thanks

PrasenjitAdhikary (approved):
LGTM

flow/flow.yaml
```diff
@@ -69,3 +69,8 @@ components:
         name:
           x-include: ../common/common.yaml#/components/schemas/Named.Object/properties/name
           x-field-uid: 7
+        payload:
+          description: |-
+            The data in the frame after protocol headers and instrumentation bytes.
```
apratimmukherjee, line 74: I think we can just keep it upto "..after protocol headers." since instrumentation bytes could be in the middle/at the end. 
Maybe : The pattern of the bytes in the transmitted frames for the flow after the protocol headers.

flow/payload.yaml
```diff
@@ -0,0 +1,48 @@
+components:
+  schemas:
+    Flow.Payload:
+      description: |-
+        A container for different types of payload, which is the data in the frame after protocol headers.
+        Some part of the payload will be overwritten with instrumentation data, contents and placement of
```
apratimmukherjee, line 6: "will be" --> "could be" ?  [ e.g. metrics not enabled so no instrumentation data inserted or other implementations might have other ways to correlate Tx/Rx information ]

flow/payload.yaml
```diff
@@ -0,0 +1,48 @@
+components:
+  schemas:
+    Flow.Payload:
+      description: |-
+        A container for different types of payload, which is the data in the frame after protocol headers.
+        Some part of the payload will be overwritten with instrumentation data, contents and placement of 
+        which could be implementation specific.
```
apratimmukherjee, line 7: which is ..

flow/payload.yaml
```diff
@@ -0,0 +1,48 @@
...
+      type: object
+      properties:
+        choice:
+          description: |-
+            A choice used to determine the type of payload.
```
apratimmukherjee, line 12: "...the pattern of the bytes in the payload following the protocol headers . " maybe.

apratimmukherjee: I think we need to document the expected behaviour of each choice  . 
e.g. fixed would insert user defined pattern in the payload bytes. 
       increment_byte would set the first byte as 00 , second as 01 , upto FF and then the pattern would be repeated. 
       decrement_byte would set the first byte as FF , second as FE , 255th as 00 and then the pattern would be repeated. 
       increment_word would set the first 4 bytes as 00 00 00 00 , second as 00 00 00 01 and so on. 
      decrement_word would set the first byte as FF FF FF FF , second as FF FF FF FE and so on.

apratimmukherjee (approved):
Left some comments to possible improve documentations and clarify expectations in some places.

apratimmukherjee (approved):
Looks good to me.

PR #443, opened 2025-09-12
Remove rectify regex patterns

Vibaswan:


apratimmukherjee (approved):
Looks good to me. 
Need to check that  [0x|0X][upper or lower case hex] works for all all platforms for the hex value with regexp.

PR #471, opened 2026-04-06
making "destination_ip_address" an optional field

satyamsinghKey:
### Feature Overview

- **Brief Description:**  
 There can be instances where we do not want to set destination Peer for a RoCEv2 peer. In that case, keeping the "destination_ip_address" a required field doesn't makes sense. Hence removing it from the required section.

### Code snippets
    rocev2_1, rocev2_2 = d1.rocev2, d2.rocev2
    rocev2_1_int, rocev2_2_int = rocev2_1.ipv4_interfaces.add(), rocev2_2.ipv4_interfaces.add()
    rocev2_1_int.ipv4_name, rocev2_2_int.ipv4_name = ip1.name, ip2.name
    rocev2_1_int.ib_mtu, rocev2_2_int.ib_mtu = 1027, 1027
    rocev2_1_peer, rocev2_2_peer = rocev2_1_int.peers.add(), rocev2_2_int.peers.add()
    rocev2_1_peer.name, rocev2_2_peer.name = "RoCEv2 1", "RoCEv2 2"
    rocev2_1_peer.destination_ip_address = ip2.address
    ###Not Setting destination ip address for 2nd peer

### DIFF
--- a/device/rocev2/rocev2ipv4.yaml
+++ b/device/rocev2/rocev2ipv4.yaml
@@ -35,7 +35,7 @@ components:
       description: >-
         Configuration for RoCEv2 IPv4 peers.
       type: object
-      required: [name, destination_ip_address]
+      required: [name]
       properties:
         name:
           x-include: ../../common/common.yaml#/components/schemas/Named.Object/properties/name
--- a/artifacts/openapi.yaml
+++ b/artifacts/openapi.yaml
@@ -12814,7 +12814,6 @@ components:
       type: object
       required:
       - name
-      - destination_ip_address
       properties:
         name:
           x-field-uid: 1
@@ -12985,7 +12984,6 @@ components:
       type: object
       required:
       - name
-      - destination_ip_address
       properties:
         name:
           x-field-uid: 1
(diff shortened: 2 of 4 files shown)


device/rocev2/rocev2ipv4.yaml
apratimmukherjee, line 44: There is an ambiguity in this change. Earlier it was required so user was forced to provide the dest IP. Now, if user does not specify, I think there is no way for an implementation to know if user actually provided it or not. 
Implementation is forced to make a is_present check which we normally never do for a attribute. This seems to be a problem as per regular design. Normally optional fields are encapsulated inside an object and we check for presence or absence of the object or some enum to drive the behaviour. 
e.g. enum dst_ip_mode = user_specified, auto_detected. 
     user_specified takes in IP , auto_detected specified how IP should be set by user / the behaviour. 
Though this does break backward compatibility as designed now.  Please take a look as such and what you want to do for snappi_ixn. 

Anyway, there should be clear documentation on expected RoCEv2 behaviour if IP is not specified. 

Same comments hold for IPv6 as well.

PR #476, opened 2026-04-27
Review isis otf on link metric

SuryyaKrJana:
### Feature Overview
- **Related Issue:** (https://partnerissuetracker.corp.google.com/issues/433281254)
- **Brief Description:**  
TO change of ISIS Link Metric on the Fly only however added other attributes and BGP OTF to oversee over design.

---
### Feature Details
 - [Redocly docs for this feature/branch](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/open-traffic-generator/models/review-isis-otf-on-link-metric/artifacts/openapi.yaml&nocors#tag/Configuration/operation/update_config)
  
**IS-IS Link Metric OTF Update — Model Change Summary**
  - Simulated interface (no real adjacency): The session stays Up. The router immediately re-advertises its LSP with the new metric value, which neighbors receive without a session interruption

-    This change introduces on-the-fly (OTF) update support for IS-IS interface link metrics via the update_config API. It adds one new file and extends one existing schema.
- New file: updateisisconfig.yaml
- UpdateProtocol.Config.Isis.Interface.UpdateGroup
- Targets a set of IS-IS interfaces for a batch of attribute updates.
         - names: string[]: of IS-IS interfaces to update. Interface names are globally unique, so no router name is required.
         - attributes: array, One or more attribute updates to apply. Multiple attributes can be updated in a single group      without repeating the names list
-  Note that Emulated interface (real adjacency with DUT): The IS-IS session will flap. The controller disables and re-enables the session; the updated metric takes effect when the session comes back up.

- [Dev-Snappi branch reference] (https://github.com/open-traffic-generator/snappi/tree/refs/heads/review-isis-otf-on-link-metric)
- Example command to fetch the dev-snappi branch:
  ```
  go get github.com/open-traffic-generator/snappi/gosnappi@<review-isis-otf-on-link-metric>
  ```
---
### Code snippets
1. Update metric on p1d1IsisIntf to 20
2. Note: this is an emulated interface (adjacency with DUT) so the IS-IS
3.   Session will flap; wait for recovery before verifying.
### Key OTF update call:

```
       cu := gosnappi.NewConfigUpdate()
       cu.Protocols().Isis().Interfaces().Add().
       SetNames([]string{"p1d1IsisIntf"}).
       Attributes().Add().SetMetric(20)
       client.UpdateConfig(cu)
```
```
{
  "choice": "protocols",
  "protocols": {
    "isis": {
      "interfaces": [
        {
          "names": ["intf-1", "intf-2"],
          "attributes": [
            { "choice": "metric", "metric": 20 }
          ]
        }
      ]
    }
  }
}

```

---


config/updateisisconfig.yaml
```diff
@@ -0,0 +1,87 @@
...
+            metric:
+              x-field-uid: 1
+          x-field-uid: 1
+        metric:
+          description: >-
+            The metric value to be updated on the IS-IS interface.
+            The maximum value of 16777215 (2^24-1) applies only when wide metric is
+            enabled (RFC 5305). With narrow metrics the maximum is 63; values above
+            63 will be rejected.
+            Updating the metric on an emulated interface will flap the IS-IS session.
```
apratimmukherjee: I think the wording and intent needs change here . 
I was looking something on these lines: 
a) If an ISIS session is in up state, the change of metric should trigger an LSP Update without bringing down the IS-IS session with its neighbor(s). 
Note : If an implementation is not able to change the metric without flapping the session in certain scenarios e.g. if metric is updated on the emulated interface ( as opposed to simulated IS-IS interface ) , it should return a warning indicating that the session will be disabled and re-enabled, and the updated attribute will be reflected once the session comes back up ) [ Similar to the language in the UpdateGroup definition ]

config/update.yaml
```diff
@@ -41,4 +47,14 @@ components:
...
           items:
             $ref: '#/components/schemas/Flow'
-          x-field-uid: 2
\ No newline at end of file
+          x-field-uid: 2
+          
+    Protocols.Update:
+      description: >-
+        A container of Protocols with associated properties to be updated that may or may not
+        able to maintain the current Protocols session in Up state always.
```
apratimmukherjee: I think the text for Warning should be moved to this level from ISIS Update Group. This is generic for the full Protocols.Update tree : 
" If the session is up but true on-the-fly update is not supported for one of the attributes selected for update  (e.g. metric change on an emulated interface): a warning should be returned indicating that the session will be disabled and re-enabled,and the updated attribute(s) will be reflected once the session comes back up."
Also, this is a good place to define all the common behaviours : 
e.g. 
a) If any of the protocol session(s) linked to the attributes is not in started state , what is the behaviour expected ? [ Error rejecting the full update request OR will be applied on next start ? ] 
b) The attributes are permanently updated in the stored config [ i.e. next get_config should return modified parameters ] 
c) If session tied to an attribute is not in started state , the change will be reflected when session comes up next .

config/updateisisconfig.yaml
```diff
@@ -0,0 +1,87 @@
...
+          x-field-uid: 1
+
+    UpdateProtocol.Config.Isis.Interface.UpdateGroup:
+      description: >-
+        An update group targeting one or more IS-IS interfaces. All interfaces
+        listed in names receive every attribute update specified in the attributes
+        list. Use multiple update groups to apply different updates to different
+        subsets of interfaces (asymmetric updates).
+
+        If the session is up but true on-the-fly update is not supported for an
```
apratimmukherjee, line 40: Maybe move this text to generic place above . Maybe also good to keep this hint for now for the metric update w.r.t. emulated interface . But probably not needed here also.

apratimmukherjee (commented):
Looks okay in terms of attibutes but needs some workflow clarifications in documentation.

config/updateisisconfig.yaml
```diff
@@ -0,0 +1,94 @@
...
+        attribute (e.g. metric change on an emulated interface): a warning is
+        returned indicating that the session will be disabled and re-enabled,
+        and the updated attribute will be reflected once the session comes back up.
+      type: object
+      properties:
+        names:
+          description: >-
```
alakendu, line 51: I feel normally we use `x-constraint`  to properly point out the interface name. So, it will good to use that.

alakendu: Changed x-constraint from double-quoted to unquoted format to fix
rendering issue in redocli UI. Double quotes caused the constraint
reference to not display properly.

SuryyaKrJana (commented):
Looks fine.

Issue #484, opened 2026-07-09
Add support for flapping BGP peers and routes


atul-nexthop:
**Is your feature request related to a problem? Please describe.**

Today OTG supports bringing BGP peers up/down and advertising/withdrawing routes as discrete operations, but there is no way to *flap* them — i.e. repeatedly toggle them over time.

**Describe the solution you'd like**

A way to flap, on the fly (while the session is up) and without changing configuration:
- **BGP peer sessions** — repeatedly bring a peer down and back up.
- **Routes** — repeatedly flap route ranges.

The capability should let a user:
- **Enable/disable** flapping on demand (start it, and stop it).
- **Configure the flap timing** — how long the peer/route stays up (advertised) and how long it stays down (withdrawn) in each cycle.
- **Select the targets** — specific BGP peers / route ranges, or all of them.

**Fields (per IxNetwork prior art)**

*BGP peer flap* — see first screenshot (peer "Flap" tab):
- **Flap** (enable/disable)
- **Uptime in Seconds**
- **Downtime in Seconds**

*Route range flap* — see second screenshot (BGP IP Route Range "Flapping" tab). In addition to enable/uptime/downtime, route ranges expose more controls:
- **Enable Flapping**
- **Uptime In Seconds**
- **Downtime In Seconds**
- **Delay in Seconds** — delay before flapping begins after it is enabled.
- **Partial Flap** — flap only a subset of the routes within the range, instead of the whole range.
- **Flap From Route Index** — start index (inclusive) of the subset to flap when Partial Flap is enabled.
- **Flap To Route Index** — end index (inclusive) of the subset to flap when Partial Flap is enabled.

**Use case**

Flapping is a common test scenario for validating convergence, reconvergence, route churn handling, and overall stability of a DUT under repeated peer/route instability.

**Reference / prior art**

This capability already exists in IxNetwork, exposed as configurable fields on BGP peers and route ranges. The peer offers an **Enable Flapping** toggle with **Up Time** / **Down Time** durations; BGP route ranges additionally expose **Delay (seconds)**, **Partial Flap**, and **Flap From / Flap To Route Index**, allowing a delayed start and flapping of only a contiguous subset of routes in the range. Including this as a reference for the expected behavior; the OTG modeling of it is of course up to the community.

_First screenshot: BGP peer **Flap** tab. Second screenshot: BGP IP Route Range **Flapping** tab._

[...]

PR #485, opened 2026-07-09
Add BGP peer and route flap control operations

atul-nexthop:
Addresses #484

## What this adds
On-the-fly *flap* operations (start/stop while the session is up, no config change), per the requirements in #484:
- **`action.protocol.bgp.peer.flap`** — flap BGP peer sessions.
- **`action.protocol.bgp.route.flap`** — flap BGP route ranges.

## Design
Flapping is only supported on **BGP** peers and route ranges in IxNetwork; ISIS/OSPF route ranges have no flap fields (only a static `Active` flag). So both flap operations are modeled **BGP-scoped** under `Action.Protocol.Bgp`, alongside `notification` and `initiate_graceful_restart`:

- **`Action.Protocol.Bgp.Peer.Flap`** — `peer_names` (BGP peers), `enable`, `up_time`, `down_time`.
- **`Action.Protocol.Bgp.Route.Flap`** — `route_names` (BGP v4/v6 route ranges), `enable`, `up_time`, `down_time`, plus the extra route-range controls IxNetwork exposes:
  - `delay` — delay (seconds) before flapping begins.
  - `partial_flap` — flap only a subset of the routes within each range.
  - `flap_from_route_index` / `flap_to_route_index` — inclusive index window for the subset when `partial_flap` is enabled.

Flap is a triggered operation with timing, which fits the *action* scope per `control/readme.md` (states are representable purely by enums; actions are triggers). `x-field-uid`s are appended so existing fields are untouched. Both new operations are marked `x-status: under_review`.

### Note on scope (changed from the earlier revision of this PR)
Route flap was initially placed on the generic `State.Protocol.Route` (cross-protocol via its `x-constraint`). That advertised a flap capability for ISIS/OSPF that the backend can't honor, and the extra fields (`partial_flap`, `flap_from/to_route_index`) are BGP-route-range specific. It has been moved to `action.protocol.bgp.route.flap` and removed from `State.Protocol.Route` (which keeps `names` + `state` withdraw/advertise, honored by BGP/ISIS/OSPF via `Active`).

## Files
- `control/bgp.yaml`, `control/protocol.yaml` (source)
- `artifacts/openapi.yaml`, `artifacts/otg.proto` (regenerated via `build.py`)

## Testing
- `python build.py` passes (openapiart validation green).

### DIFF
--- a/artifacts/openapi.yaml
+++ b/artifacts/openapi.yaml
@@ -20162,40 +20162,6 @@ components:
           enum:
           - withdraw
           - advertise
-        flap:
-          x-status:
-            status: under_review
-            information: This is subject to change.
-          $ref: '#/components/schemas/State.Protocol.Route.Flap'
-          x-field-uid: 3
-          description: |-
-            Under Review: This is subject to change.
-
-            Description TBD
-    State.Protocol.Route.Flap:
-      description: |-
-        Flaps the selected routes on the fly by continuously withdrawing and re-advertising them.
-      type: object
-      required:
-      - enable
-      properties:
-        enable:
-          description: |-
-            Enables or disables flapping of the selected routes.
-          type: boolean
-          x-field-uid: 1
-        up_time:
-          description: |-
-            The duration (in seconds) for which each route stays advertised in a flap cycle.
-          type: integer
-          format: uint32
-          x-field-uid: 2
-        down_time:
-          description: |-
-            The duration (in seconds) for which each route stays withdrawn in a flap cycle.
-          type: integer
-          format: uint32
-          x-field-uid: 3
     State.Protocol.Lacp:
       description: |-
         Sets state of configured LACP
@@ -20998,11 +20964,14 @@ components:
               x-field-uid: 2
             peer:
               x-field-uid: 3
+            route:
+              x-field-uid: 4
           x-field-uid: 1
           enum:
           - notification
           - initiate_graceful_restart
           - peer
+          - route
         notification:
           $ref: '#/components/schemas/Action.Protocol.Bgp.Notification'
           x-field-uid: 2
@@ -21015,6 +20984,16 @@ components:
             information: This is subject to change.
           $ref: '#/components/schemas/Action.Protocol.Bgp.Peer'
           x-field-uid: 4
+          description: |-
+            Under Review: This is subject to change.
+
+            Description TBD
+        route:
+          x-status:
+            status: under_review
+            information: This is subject to change.
+          $ref: '#/components/schemas/Action.Protocol.Bgp.Route'
+          x-field-uid: 5
           description: |-
             Under Review: This is subject to change.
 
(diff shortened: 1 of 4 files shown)


atul-nexthop:
@apratimmukherjee @sasubrata this implements #484 — on-the-fly flap for BGP peers and routes. I've raised it as a proposal and I'm happy to adjust the modeling to fit the project's conventions. Could you take a look or assign a reviewer when you have a chance? Thanks!

apratimmukherjee:
@atul-nexthop Thanks for raising this — appreciate the initiative on both the issue and the PR.

I want to check this against a related ask: I was forwarded an email thread describing a need to not auto-advertise routes on session up, and instead control route transmission (and flapping) explicitly from the test.

Could you clarify whether this PR is:

An additional requirement — i.e., flap control exposed via the API itself, rather than driven from test code (e.g., a loop issuing adv/withdraw), or 
Related to the same need described in that email thread?
That'll help us figure out whether this should land together with the session-bringup/route-advertisement separation work, or as its own item — will follow up once we hear back. Thanks!

atul-nexthop:
> I want to check this against a related ask: I was forwarded an email thread describing a need to not auto-advertise routes on session up, and instead control route transmission (and flapping) explicitly from the test.

@apratimmukherjee  Thank you for your quick response.

This is separate task from the one raised via support team.


- So the first ask which was raised via support team, It was ability to bifurcation of the route advertisement or withdrawl from BGP session establishment - currently, The routes are auto advertised when session comes up - The specific use for this is to allow us to measure route programing learning capability.
- The ask in this PR is separate - We would like the ability to set the flap and timers via api - If you refer to the screenshot attached on this Issue description - IXIA already has flap parameter in BGP Peer and BGP route ranges - so, This PR is to allow us ability to set those so specific routes are only flapped allowing us better control over flap timers as well.

Please let me know if this makes sense.

atul-nexthop:
@apratimmukherjee @indraniBh @kamalsahu0001  I would appreciate your help on this if these fields can be exposed on OTG and allows us to set these using snappi.

if you would like me to open a support ticket - please let me know

indraniBh:
@atul-nexthop :
We are working on a script-based approach and will share an update soon.

One point worth highlighting is that the background thread mechanism would only be required if this verification needs to run asynchronously in the background. Otherwise, the same behavior can be implemented in a straightforward manner as a direct library function/wrapper.

We are leaning toward a library-based solution rather than introducing a new OTG API for a couple of reasons:
1. Keeping OTG vendor-neutral and the API surface simple:
OTG aims to remain vendor-neutral and provide an extensible toolkit. In general, if a requirement can be addressed through a library wrapper built on top of existing APIs, we prefer that approach over adding new protocol-specific APIs. This helps keep the core API surface easier to understand, implement, and maintain across vendors.
2. Avoiding overlap with existing capabilities:
In this case, introducing a dedicated API could potentially overlap with existing route advertise/withdraw APIs as well as APIs already available for stopping/restarting specific BGP peers. Those primitives can be combined as needed and triggered based on custom verification logic implemented in the wrapper layer.
An additional benefit of this approach is that it remains generic and can be applied to other protocols (for example, OSPFv2/v3 and ISIS) that may require similar verification workflows, without introducing protocol-specific APIs or requiring additional implementation and validation effort across all supported implementations.

We are putting together this approach and will update the thread shortly.

atul-nexthop:
Thanks @indraniBh — and I appreciate you working up a concrete approach. I agree with both goals: keep OTG vendor-neutral, and don't add API surface where existing primitives compose. Before the script lands, two things I'd ask you to weigh, because I don't think a wrapper can cover them.

**1. Call volume, and the robustness that follows from it.** With engine-driven flap the test makes **two calls total** — start with the up/down timers, stop when done — and the chassis runs the pattern itself. A wrapper makes **two calls per flap cycle**, so the count scales with duration and flap rate: a 10-minute run at a 1-second period is ~1,200 REST calls instead of 2, and a longer or faster test multiplies from there.

That has two consequences. The obvious one is load — those calls land on a shared API server that other sessions and other teams are using at the same time. The one that worries me more is fidelity: every transition depends on that path staying healthy for the whole run, so the flap pattern degrades exactly when the controller is busiest, which on a scale test is precisely when the flap is running. The stimulus the test asked for stops being the stimulus the DUT sees, and a flap test is only meaningful if the stimulus is known and repeatable. Is that something the wrapper approach can guarantee?

**2. Partial flap.** `partial_flap` with `flap_from_route_index`/`flap_to_route_index` flaps a subset *within* a route range. Advertise/withdraw works at range granularity, so a wrapper would have to split the range into separate ranges — which changes the configuration under test and invalidates what's being measured.

**On genericity — I think we want the same thing.** An earlier revision of this PR put flap on the generic `State.Protocol.Route`, cross-protocol. I moved it to BGP scope only because ISIS/OSPF expose a static `Active` flag and can't honour flap, and I didn't want to advertise a capability backends can't deliver.

If the concern is protocol-specific surface, I'm happy to move it back to a **generic route/protocol scope with capability reporting**, so unsupported backends report unsupported rather than silently no-op. That gives one operation instead of per-protocol APIs, addresses the overlap concern, and keeps the engine-side execution a wrapper can't provide.

Happy to revise the PR that way, or to review your script against these two cases once it's up.

